PR #82, opened 2014-04-25
Turn off doclint support during JavaDoc builds.

irbull:
In Java8 doclint is enabled by default. The existing JavaDocs
do not properly compile with this setting, and the build fails
on Java8. This fixes the problem by turning doclink off
in the top level POM.xml

A maven profile was used so this still builds on older JDKs
that don't understand this flag.

For more information see [1,2]

[1] http://mail.openjdk.java.net/pipermail/build-infra-dev/2013-January/002899.html
[2] http://stackoverflow.com/questions/15886209/maven-is-not-working-in-java-8-when-javadoc-tags-are-incomplete

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,25 @@
 			</repositories>
 		</profile>
 
+		<!-- SPECIAL PROFILE FOR JAVA 8-->
+		<profile>
+			<id>doclint-java8-disable</id>
+			<activation>
+				<jdk>[1.8,</jdk>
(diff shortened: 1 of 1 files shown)


savageautomate:
Thank you for the contribution.

savageautomate:
Moved changeset into develop branch:
https://github.com/Pi4J/pi4j/commit/52fa51e23557ff14f064300484d2e147d8dfe747

PR #88, opened 2014-06-12
Added package com.pi4j.temperature to the set exported in the bundle manifest

lathil:
I am using pi4j throught a osgi framework. Unfortunately, the package com.pi4j.temperature doesn't get exported. Could it be possible to add it ?


savageautomate:
merge complete in develop branch

Issue #85, opened 2014-05-21
Tomcat deploy and "java.lang.UnsatisfiedLinkError: com.pi4j.wiringpi.Serial.serialOpen"


cacodemon79:
Hi friends,
I have a strange problem with Tomcat 7 and Pi4J.

I have a very simple servlet that prints on a web page the data read from the serial port (after 10 seconds).

I deploy WAR trough ANT build.xml on a remote TOMCAT 7 (installed in a RASPBERRY PI environment).

When i start TOMCAT for the first time, all works very well. Serial data are read and written on a web page.

May 21, 2014 10:27:19 PM org.apache.coyote.AbstractProtocol pause
INFO: Pausing ProtocolHandler ["http-bio-8080"]
May 21, 2014 10:27:19 PM org.apache.catalina.core.StandardService stopInternal
INFO: Stopping service Catalina
May 21, 2014 10:27:20 PM org.apache.coyote.AbstractProtocol stop
INFO: Stopping ProtocolHandler ["http-bio-8080"]
May 21, 2014 10:27:20 PM org.apache.coyote.AbstractProtocol destroy
INFO: Destroying ProtocolHandler ["http-bio-8080"]
May 21, 2014 10:27:40 PM org.apache.coyote.AbstractProtocol init
INFO: Initializing ProtocolHandler ["http-bio-8080"]
May 21, 2014 10:27:41 PM org.apache.catalina.startup.Catalina load
INFO: Initialization processed in 14641 ms
May 21, 2014 10:27:42 PM org.apache.catalina.core.StandardService startInternal
INFO: Starting service Catalina
May 21, 2014 10:27:42 PM org.apache.catalina.core.StandardEngine startInternal
INFO: Starting Servlet Engine: Apache Tomcat/7.0.28
May 21, 2014 10:27:42 PM org.apache.catalina.startup.HostConfig deployDescriptor
INFO: Deploying configuration descriptor /etc/tomcat7/Catalina/localhost/host-manager.xml
May 21, 2014 10:27:52 PM org.apache.catalina.util.SessionIdGenerator createSecureRandom
INFO: Creation of SecureRandom instance for session ID generation using [SHA1PRNG] took [125] milliseconds.
May 21, 2014 10:27:52 PM org.apache.catalina.util.SessionIdGenerator createSecureRandom
INFO: Creation of SecureRandom instance for session ID generation using [SHA1PRNG] took [125] milliseconds.
[...]

PR #100, opened 2014-09-22
 #85 reimplementing NativeLibraryLoader

PoJD:
- simplified, only loading from embeded resource, trying both soft and
  hard versions
- fixes issues with redeployment in J2EE environments, since this method
  now effectively creates new so temp file each time it is invoked. All
  of the temp files then get deleted on web container shutdown
- dropped library name from the method too, only using the filename of
  the so library to load
- updated all client code
- eclipse seem to have updated many other comments and copyrights with
  current year too
- added more eclipse files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ maven-eclipse.xml
 .externalToolBuilders
 /target
 /.settings
+.metadata/
+RemoteSystemsTempFiles/
+
 /pi4j-core/.settings
 /pi4j-core/.classpath
 /pi4j-example/.settings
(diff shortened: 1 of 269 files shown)


PoJD:
Please do not merge yet, I just found out my local tomcat runs at 100%CPU after redeploy. FIrst deploy is fine. But not sure whether it is caused by this change or by some other changes I did locally in my app. Will debug more and try to dig into. Feel free to play around/comment as well, thanks!

savageautomate:
Also, please make sure you are making changes to the 'develop' branch and not the master branch.  This is where the (not-yet-released) current 1.0-SNAPSHOT code lives.    Support for soft-float has already been dropped in the latest loader code.

PoJD:
Ah, OK, sorry, missed that. Will do.

PR #101, opened 2014-09-23
Reimplementing loading native library

PoJD:


PoJD:
I can still see tomcat runs to 100% CPU after hot deploy. I tried with released version 0.0.5, which does not even work after hot deploy (gives UnsatisifedLinkErrors) and the same CPU 100% is there. So either this is something else in my app (JDBC, etc) or something else inside Pi4J directly. At least this pull request refactors the loading of native libraries and hot deploy now works fully - everything is loaded again, new so lib for JVM to load. I call unexportAll, unprovisionAll and shutdown on GPIO on undeploy of webapp to clean as much as I can. I will try to troubleshoot the CPU issue, if I get to the bottom and it turns out to be Pi4J (I doubt it), then I may raise another issue here

PoJD:
OK, tried using my custom Mock GPIO in my webapp and after redeploy the CPU is at 0% as usual. SO this turns out to be something inside Pi4J. But since the behavior is the same before my change and after it, it is unrelated to this. I will try digging further later and potentially raise a new issue for that here.

PoJD:
Actually it could very well still be something else than Pi4J since there is other logic in my webapp that only runs when there is real GPIO and valid readings are coming back from the sensors. Is there anything I should do now for this pull request to get merged? Thanks!

PoJD:
OK, solved my 100% CPU issue after undeploy. It was in the way I did cleanup of GPIO in my app. Obviously I did unexport all before shutdown which confused things. I suspect some of the threads may still be running when doing unexport and therefore were endlessly polling GPIO. Since I did shutdown till after unprovision all, it made things even worse since all pins were already removed from the internal collection and therefore the polling could not be fixed (looking inside shutdown I can see it loops through the pins too). Now proper order in my app is to shutdown (using shutdown options, so no need to call unexportAll explicitelly) and then unprovision all to completely cleanup and all is OK. Just FYI, it did not really have any impact on this pull request due to the above, but just to provide the transparency...

https://github.com/PoJD/hawa/commit/4eb9873cfbc47b75cecea8dbf3ad3f9fd2828cf7

savageautomate:
I have not been able to review this changeset yet, but will hopefully will next week.  Thanks!

PoJD:
Thank you!

PoJD:
Hello, any luck? :)

PoJD:
Hello, any update please? I have one more pull request #112 ready and I had to merge this change already locally since I need all these fixes in. I would love to continue contributing back to this repository as opposed to continue my work in the fork only! Thanks

savageautomate:
I am reviewing and merging today; sorry for the delay.

savageautomate:
I'm a little concerned that removing the native system loader logic where it would first attempt to load the library from the existing classpath or manually configured library path may negatively impact some users?  While this was not the default path for loading the native .SO file, it did provide an alternative method to manually copy the .SO file to a system libs path or allow users to include the command line argument "-Djava.library.path=" to manually specify a path.  Maybe my concerns are theoretical so I'm willing to merge this change and see if it impacts anyone in a negative way.  If it does we may need to re-include some ability to load the native lib from a configured path (perhaps in a friendlier way that does not rely on exceptions).

Thanks for this work!

PoJD:
Thanks!
Dne 28. 11. 2014 16:45 "Robert Savage" notifications@github.com napsal(a):

PoJD:
As for the alternative loading. This is still using standard java getResourceAsStream. That looks for files in classpath. Then the order of elements in the classpath is used to determine the precedence. So nothing is dropped really - if someone needs to use a specific .so library in the system as opposed to the one embeded in the pi4j jar, all one needs to do is to add that path with the alternative .so to the classpath of the application before the jar of pi4j. I think it is still worth the simplification in the loader logic and most imporantly - it now works in JEE world too! Thanks for merging.

savageautomate:
FYI, I'll build and upload the latest snapshot later today.

Issue #17, opened 2012-12-15
Logging


savageautomate:
Instrument Pi4J with logging.
Probably use SLF4J as logging interface.

savageautomate:
**Closed**; SLF4J is supported in Pi4J v2.x

PR #112, opened 2014-11-22
Feature/spifix

PoJD:
This also includes the other pull request, I had to already merge it locally in my fork to be able to work, am still waiting for #101 to be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ bin
 .classpath
 .project
 .externalToolBuilders
+.metadata
 maven-eclipse.xml
 
 # Vim files
--- a/pi4j-core/src/main/java/com/pi4j/jni/I2C.java
+++ b/pi4j-core/src/main/java/com/pi4j/jni/I2C.java
@@ -52,7 +52,7 @@ public class I2C {
 
     static {
         // Load the platform library
-        NativeLibraryLoader.load("pi4j", "libpi4j.so");
+        NativeLibraryLoader.load("libpi4j.so");
     }
 
     /**
--- a/pi4j-core/src/main/java/com/pi4j/util/NativeLibraryLoader.java
+++ b/pi4j-core/src/main/java/com/pi4j/util/NativeLibraryLoader.java
@@ -26,11 +26,16 @@ package com.pi4j.util;
  * #L%
  */
 
-import java.io.*;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.FileHandler;
 import java.util.logging.Level;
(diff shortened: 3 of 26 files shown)


PoJD:
I am also thinking about using more OOP approach on Spi. Look at my SpeDevice abstraction in my repository: https://github.com/PoJD/hawa/blob/master/rpi/src/main/java/cz/pojd/rpi/sensors/spi/SpiDevice.java . Maybe I can raise one more pull request to get it merged to Pi4j?

PoJD:
I will update this PR at home given the other PR is already merged to develop now. Robert, let me please know your thoughts. The current SPI implementation does not work for my code for MCP3008. Or any SPI chip that needs values > 127.  More described in the commit. One thing I realize now is that this is not backwards compatible. So maybe I can keep the old method in and provide this new correctly handlingthe datatypes (unsigned char in c++ and short in java to handle it with no data loss). Also please let me know about the more OOP approach I suggested above, am happyto port some of my code here. In fact this SPI API is really more of a C++ API...

savageautomate:
OK, I have been digging into this code and I do agree that we should keep the older methods for backwards compatibility.  I think we could also expose the new (short) methods as utility methods making it easier for users to assign values and only have a single underlying JNI native method implementation.  I think we can use the existing JNI method without any modifications.  The Java unsigned byte gets turned into a jByte in JNI and then we simply iterate over the array and cast each jByte into an unsigned char.  Now in my code, I did not have an explicit cast defined, maybe that is needed?  

I'm working on the change and I'll commit it for you to take a looks at and see if it yields the same net results.  

I do like the idea of including an OO friendly SPI interface and I think your existing version could serve as a great starting point.   

I'll post back shortly.

PoJD:
Robert, this does not work for one simple reason. Java byte is signed, not
unsigned. So java byte can hold values -128 to 127. Whereas unsigned char
in c++ is unsigned, I.e. holds values 0-255. So the original implementation
works if the java client only needs to send and receive values 0-127.

In my case this was not enough. In order to talk to MCP3008, I need to send
3 unsigned chars, second of which is in a form like xxxx0000. E.g.
11110000. Which is 240 in decimal.

There is no such concept as unsigned primitive values in java, therefore we
need to use the closest higher type, which is short.

Hope this makes sense. I am happy to take the old API back in, we can have
just one native method, but we have to pass shorts at the lowest level.
 Dne 28. 11. 2014 19:30 "Robert Savage" notifications@github.com
napsal(a):

savageautomate:
OK, here is a first pass:  https://github.com/Pi4J/pi4j/commit/b4a86605f7076aa2f0f8ccc42f1adea98f19a485

I went ahead and did add a new JNI method (based on your code) for the short array.  I started using a pure Java conversion loop, but recognized that it was just wasting CPU cycles and decided just to add the new JNI method :-) 
https://github.com/Pi4J/pi4j/commit/b4a86605f7076aa2f0f8ccc42f1adea98f19a485#diff-75997c356de08ae207a327b31ca76efdR90

I left the 'length' argument in place for the new short array method:
https://github.com/Pi4J/pi4j/commit/b4a86605f7076aa2f0f8ccc42f1adea98f19a485#diff-e58ae43e6ba20c3bc55d13e6b81746fdR201
but I did add convenience methods that eliminate the length argument for both the short array and byte array methods
https://github.com/Pi4J/pi4j/commit/b4a86605f7076aa2f0f8ccc42f1adea98f19a485#diff-e58ae43e6ba20c3bc55d13e6b81746fdR177
https://github.com/Pi4J/pi4j/commit/b4a86605f7076aa2f0f8ccc42f1adea98f19a485#diff-e58ae43e6ba20c3bc55d13e6b81746fdR221

I did not return the short array -- my reasoning here is to honor the WiringPi (C) API as close as possible for this wrapper class impl.   Also, the return resulting bytes are already included in the data argument passed into the method call.  The data buffer is updated directly in the JNI code.  

I think the more conventional Java OO approach would be to return the resulting bytes from the method call, so let's do that in the OO SPI class that we create.  I'm going to mock up the interface now based on your SPI wrapper impl.  More to come ...   Thanks, Robert

savageautomate:
I understand that Java prevents you from directly assigning a value greater than 127 into the signed byte primitive type.  But the bits contained in the primitive's 8 bit storage should be correct.  For example if you assign a byte using "(byte) 0xFF" then the bits are "11111111" in the 8 bit storage.  So when this gets passed into JNI as a jByte and then cast into an unsigned char, it should cast correctly.  I'll run a test here to verify.

PoJD:
While I agree with the fact we still need 8 bits here, the assumption is wrong. http://www.myhowto.org/java/60-understanding-the-primitive-numeric-type-conversions-in-java/ explains this nicely. In simple terms java drops the most significant bits when converting, so you would actually get a different representation - also the very first bit is kept for the sign, so the above 11111111 would be interpretet as a negative number. So when java would then pass thie value down, it would treat that as the negative number. E.g. example:

```
    byte bb = (byte)0xFF;
System.out.println("b: " + bb + ", binary: " + s(bb, 8));
System.out.println("b2: " + (short)bb + ", binary: " + s((short)bb, 8));

public static String s(int number, int bits) {
return String.format("%" + bits + "s", Integer.toBinaryString(number)).replace(' ', '0');
}
```

So the above would still be -1. How would I translate that? If I would get 0xFF from the underlying C++ code, java would drop the most significant bit and treat that wrongly. I tried playing around a bit, maybe you can find some trick how to pass such values through. But even if we did, it would still mean we would need some conversion functions for the client code that would need to somehow translate to byte. Or would you expect my client code to only cast to byte and then internally transform that properly?

As for the API, I understand the honoring of wiringPi. The return sounded again more java friendly though, since in java primitive arguments to methods are never changed within the methods. Arrays are, but I dislike that API, sounds really more C++-ish :) But your call really, if you manage to get it working. The OOP wrapper would be nice though anyway and more appropriate for OOP language we use here = java.

PoJD:
Actually simple & with 0xFF would do the trick:) http://stackoverflow.com/questions/7401550/how-to-convert-int-to-unsigned-byte-and-back

So maybe we can keep only byte API in assuming you get correct values through JNI too...

PoJD:
But this still polutes the client code. If we wrap this & 0xFF inside the Spi class for the input, we would need to ask the caller to the same operation on the output. Or accept bytes and return shorts. None of which seems nice. So maybe safest and best and cleanest really to use shorts...

savageautomate:
Compiling for a test now.  I agree if any conversion is required, I'd rather just do shorts.

savageautomate:
In your example:

```
    byte bb = (byte)0xFF;
System.out.println("b: " + bb + ", binary: " + s(bb, 8));
System.out.println("b2: " + (short)bb + ", binary: " + s((short)bb, 8));

public static String s(int number, int bits) {
return String.format("%" + bits + "s", Integer.toBinaryString(number)).replace(' ', '0');
}
```

when "s(bb, 8)" is called, the byte is getting _auto-boxed_ into an int.  Thus a _conversion_ between the primitive byte and primitive integer is occurring which is attempting to honor the sign on the primitive value.  
https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

In the SPI impl case, we are never converting the byte into any other type.  We are passing it into JNI and then we are performing a cast into a native unsigned char.  This native casting does not honor any signs it simply takes the 8 bits and pushes them into the unsigned char.

savageautomate:
So here is my example code trying to figure out if my assumption is right or wrong :-)
In the Spi.java class, I added the following two methods:

```
public static native void test1(byte[] data, int length);    
public static native void test2(short[] data, int length);
```

Next the JNI native impl methods:

```
/*
 * Class:     com_pi4j_wiringpi_Spi
 * Method:    test1
 * Signature: ([BI)V
 */
JNIEXPORT void JNICALL Java_com_pi4j_wiringpi_Spi_test1
  (JNIEnv *env, jclass class, jbyteArray data, jint length)
{
    int i;
    unsigned char buffer[2048];
    jbyte *body = (*env)->GetByteArrayElements(env, data, 0);
    for (i = 0; i < length; i++) {
        buffer[i] = (unsigned char)body[i];
        printf("%u\n",buffer[i]);
    }
    (*env)->ReleaseByteArrayElements(env, data, body, 0);
}

/*
 * Class:     com_pi4j_wiringpi_Spi
 * Method:    test2
 * Signature: ([SI)V
 */
JNIEXPORT void JNICALL Java_com_pi4j_wiringpi_Spi_test2
  (JNIEnv *env, jclass class, jshortArray data, jint length)
{
    int i;
    unsigned char buffer[2048];
    jshort *body = (*env)->GetShortArrayElements(env, data, 0);
    for (i = 0; i < length; i++) {
        buffer[i] = (unsigned char)body[i];
        printf("%u\n",buffer[i]);
    }
    (*env)->ReleaseShortArrayElements(env, data, body, 0);
}
```

and finally a test snippet of code where I pass in an array of bytes and then an array of shorts with the same exact values. 

```
    System.out.println("--TESTING BYTE ARRAY VALUES --");

    byte[] bytes = new byte[5];
    bytes[0] = (byte) 0x00;
    bytes[1] = (byte) 0x0D;
    bytes[2] = (byte) 0xA1;
    bytes[3] = (byte) 0xDE;
    bytes[4] = (byte) 0xFF;

    com.pi4j.wiringpi.Spi.test1(bytes, 5);

    System.out.println("--TESTING SHORT ARRAY VALUES --");

    short[] shorts = new short[5];
    shorts[0] = (short) 0x00;
    shorts[1] = (short) 0x0D;
    shorts[2] = (short) 0xA1;
    shorts[3] = (short) 0xDE;
    shorts[4] = (short) 0xFF;

    com.pi4j.wiringpi.Spi.test2(shorts, 5);
```

The output from the JNI methods (printing the value of each unsigned char after the cast) prints the following

```
 --TESTING BYTE ARRAY VALUES --
 0
 13
 161
 222
 255

 --TESTING SHORT ARRAY VALUES --
 0
 13
 161
 222
 255
```

So this suggests that the unsigned char array in C contains the same raw values no matter if I cast from jByte or jShort.  Of course this example would not work for any values that are more than 8 bits -- but that's not our case.

PoJD:
Perfect! My example is wrong, I realized later too and commented, the simple & with 0xFF would translate the bytes to unsigned values (still 8bit of course). So you are right that we can freely pass in bytes then with no harm. The issue with the API still stands though. While the client of Pi4j can simply pass in bytes (so in the client code cast, i.e. narrow down), if we keep the output of the method in bytes (which is implicit if we follow the wiringPi API), then we are simply forcing clients to invoke code like the following:

```
byte[] bytes = new byte[3];
bytes[0] = (byte) 0x00;
bytes[1] = (byte) 0x0D;
bytes[2] = (byte) 0xA1;

com.pi4j.wiringpi.Spi.test1(bytes, 3);
short[] result = new short[3];
result[0] = bytes[0]&0xFF;
result[1] = bytes[1]&0xFF;
result[2] = bytes[2]&0xFF;
```

In other words we polute the client code with the above. Otherwise simple cast to short would yield wrong results (negative values for bytes > 127). So either we introduce the signature I suggested (returning another array) and use shorts there or use shorts only (I prefer this one).

I am still no convinced on the simple int return type, but that is your call really - you own it:) But that API really looks C++-ish to me :) But maybe the OOP approach would come to rescue then.

savageautomate:
The current commits include support for both byte[] array and short[] array.  Am I missing something?  With this the consumer should not be forced to perform data type conversions?  (If you choose to pass in an array of shorts, that array gets updated with the result values.)  I did add the second JNI method to support shorts -- no Java data type conversions! 

I agree with your sentiment on the C++-ish style of the int return and returning out data on the param.  However, these methods are supposed to just be direct ports of the WiringPi methods.  That being said, the Pi4J interfaces are where we add the special sauce (OO style) and make it easier for Java devs to consume.  We try to encourage and promote use of the Pi4J interfaces.  these WiringPi wrapper methods are just here for legacy support and cases such as the Spi where we had not yet added new interfaces.   On that note, check out this commit.  https://github.com/Pi4J/pi4j/commit/e4ad14f19981fca00f7171278ae0e174b111b7a6   Here is a start for a SPI interface and impl.  It's somewhat based on your example mixed with some existing patterns used in Pi4J.  The methods are not yet implemented but take a look at the interfaces and see what you think.

PoJD:
Apols, I missed that, I missinterpreted your post before as a proof that the byte[] API is enough. But if you keep the short[] version in, then perfect and that would suite my needs indeed and would allow people to use the API with no need for a conversion.

Thanks for such quick turnaround! I will look at the other commit and comment there what I think, thanks for asking! :)

savageautomate:
No worries :-)  Lots of stuff in this thread.  I just wanted to keep the byte[] impl in place and make sure it was working properly.  Adding the short[] impl makes perfect sense to avoid casting values larger than 127.  Originally I was thinking of converting the short[] array to a byte[] array to avoid adding a new a JNI method.  But I changed my mind to reduce array copying and avoid any type conversions in java.

PoJD:
Btw since I noticed people do contribute with other sensors etc. implementations - would you be interested in porting any of the sensors in my rpi module from my hawa repository? https://github.com/PoJD/hawa/tree/master/rpi/src/main/java/cz/pojd/rpi/sensors, the list of supported sensors there is in my git repo main page: https://github.com/PoJD/hawa . I am happy to port it (not sure about the sensor API, maybe that is too specific for my app, but am happy to meet any API you would like me to).

savageautomate:
Yes, I think many of these would be great additions to Pi4J.  We may not have the appropriate component interfaces for some of the sensors, but we could easily add them.  

The AtoD (MCP3008) implementations would translate to "GPIO Extensions" included in the Pi4J gpio-extension library
https://github.com/Pi4J/pi4j/tree/develop/pi4j-gpio-extension/src/main/java/com/pi4j/gpio/extension

> Example: MCP4725 D2A impl:
> https://github.com/Pi4J/pi4j/tree/develop/pi4j-gpio-extension/src/main/java/com/pi4j/gpio/extension/mcp
> 
> Example: ADS1015, ADS1115 A2D impl:
> https://github.com/Pi4J/pi4j/tree/develop/pi4j-gpio-extension/src/main/java/com/pi4j/gpio/extension/ads

The light (lux), barometric pressure, temperature, humidity etc. sensors would be ideal candidates to include the the Pi4J device/component library.
https://github.com/Pi4J/pi4j/tree/develop/pi4j-device/src/main/java/com/pi4j/component

If you have the time to port any of these in, that would be a great help!
Thanks, Robert

savageautomate:
Closing pull request, I think we have included the core intentions of this PR in the latest commits (independant of this changeset).

PoJD:
Excellent, will first try your latest builds and then start porting over in upcoming weeks.Do you have any preference for cases where a native code is needed? E.g. the humidity sensor is using a native code (little adoption of adafruit c libraries).

savageautomate:
Currently all native code lives in the "pi4j-native" project.  This makes it nice for the build to isolate native compiling to that one project.  However, only "pi4j-core" currently consumes the native lib.  So maybe some trial and error ahead.

PoJD:
Fair enough. I did build pi4j-native locally before (when I was playing
with my custom changes to the SPI API), so am familiar with that. So will
probably make a new dependency on the this native in the extension project
and see how it goes:) Will ping you back if I face some issues, thanks!

PoJD:
I have started porting, tried with MCP3008 for now: #115 - let me please know what you think about that, I did a few things differently there (e.g. used enum rather than pin constants, added jmockit, commons logging, etc). About the enum - ideally we would make somehow more strict exporting (e.g. allow exporting only pins of certain types for providers, maybe we can do that later. At least the enums seemed like less code to me now and probably more readable. But I can easilly change that to match the rest of the providers.

savageautomate:
I merged the changeset with a few modifications.  

I removed the logging impl. -- _not because I hate logging_ -- I wish all of the library included logging (see feature request #17) -- I just think we need to define a single logger interface (commons-logging, java util logging, or SLF4J) for the whole project and consistently implement it everywhere.

I went ahead and converted the enum for the pins back to static class members.  I liked you impl and reasoning, but this seems more consistent with all the existing impls.

savageautomate:
I have a MCP3008 chip on order so I'll get to try this out for myself soon!

savageautomate:
BTW, can you contact me via email ....

PoJD:
Hi Robert, email sent. Thanks for the merge! I vote for commons logging simply because I have always used it and also because it is already used in my app, so it would be easier for me to integrate:) But I appreciate the latter is not really any argument for you:) I was also thinking about adding those deps to root pom. Am happy to take it forward. We can maybe only leave commons logging in and do not provide any lower level logging(log4j in my case), thus leaving it to the client. Am also happy to try out some refactoring about the pins (enum rather than fields) everywhere and could try to incorporate it in the rest of the code to provide more safe API. It actually happened to me too, that I used MCP23017 provider, but with wrong pins! We might need to leave the current API still in though for backward compatibility, we would need to think it over.

savageautomate:
I have always used slf4j as my logging interface at work.  But I was thinking about looking at Java util logging since those interfaces don't require any additional project dependencies.  I'm not sure what the trade offs are yet.  At the moment Pi4J does not have any external dependencies thus a consumer can simply import the Pi4J core jar by itself and get up and running with minimal effort.  I think this is an attractive feature for a library, but there have been many times where I really wanted some apache commons libs :-)

PR #115, opened 2014-11-30
MCP3008 provider

PoJD:
- added new provider for MCP3008 10bit ADC
- added test cases using jmockit
- adding new maven dependencies to jmockit, junit, log4j and commons
  logging
- added test resource for log4j
- introducing the pins as enumeration to have more safe way to assign
  PINs


PoJD:
Btw I tested that with my app and it works fine. https://github.com/PoJD/hawa/commit/d4be92ace9fb144f8861538dc935500d265e9aca

PR #137, opened 2015-03-05
Added MCP44xx, MCP45xx and MCP46xx families to pi4j-i2c-devices

RasPelikan:
As announced in pi4j-google group (https://groups.google.com/forum/#!topic/pi4j/mMQjb7WKrlY) I'm now ready for the pull-request!

All 24 devices of microchip's 44xx, 45xx and 46xx families are compatible in I2C communication. I defined all of them according their specifications. Additionally I added a device-interface pi4j-device/com.pi4j.device.potentiometer.DigitalPotentiometer which is implemented by all classes in pi4j-i2c-devices/com.pi4j.i2c.devices.microchip.potentiometers.

Real-life-testing is done by using the ICs 4641 (EEPROM) and 4651 (RAM). I will write a blog-post soon which will document the tests by videos.

I wrote a lot of detailed unit-tests, so I hope the quality is sufficient.

Greetings from Vienna/Austria,
Stephan


savageautomate:
Hi Stephan,

I am going to refactor your changes into the pi4j components namespace (in the pi4j device project).  I think these are a better fit for the components library.  Currently the pi4j-i2c-device project is not distributed with the pi4j builds whereas the pi4j-device.jar is.  

Thanks, Robert

RasPelikan:
Fine!

I built pi4j-i2c-devices seperatly, copied the jars into distribution/target and run "mvn package". I'm looking forward to your refactoring!

Stephan

savageautomate:
Hi Stephan,

First, let me say ... excellent work on these and thank you for the contribution.  Certainly looks very professional!

Take a look at the refactoring changes here:
https://github.com/Pi4J/pi4j/commit/8a03d42a03e3310eb17e7d240170f504afd709da

Most of the classes ended up here:
https://github.com/Pi4J/pi4j/tree/develop/pi4j-device/src/main/java/com/pi4j/component/potentiometer

I tried to follow the conventions of pi4j's components as close as possible; however there is one negative side-effect ... some of the class names are now a bit verbose.  Functional .. just lengthy.   I'm not married to the names, so let me know what you think.

Basically the idea for components are reusable implementations of electrical components that adhere to a unified component interface.   So it seemed like 'Potentiometer' was a perfect fit alongside existing components such as Relay, Light, LED, Motor, etc.   So components are these low level things that represent some hardware component implementation.  There is also a "device" namespace in the 'pi4j-device' project and that is intended for some device or system that may be a composite collection of components like a sprinkler system or garage door opener.   

The 'pi4j-i2c-device' project was added as somewhat of a placeholder/playground for some work that was started.  After looking a little deeper into it some of of these existing impls also need move to components and some are incomplete.   I may rename the project to "pi4j-sandbox" or something to help avoid any confusion.  

Thanks again for your work!  Let me know what you think about the refactoring -- I'm very open to any changes.   

Thanks, Robert

RasPelikan:
Hi Robert,

thank you for the flowers, but I have to say "thank you!" for providing the project!

> so let me know what you think.

Naming is always difficult. What I try is to image what the user expects on using my classes. I will complete this thought after commenting the next point of you:

> Basically the idea for components are reusable implementations of
> electrical components that adhere to a unified component interface

This is called loose coupling. The idea is to unbind from specifics. So indeed my tests would be better if they would use

```
private DigitalPotentiometer potiA;
```

instead of

```
 private TestableMCP45xxMCP46xxPotentiometer potiA;
```

Also my test-main-classes in MCP4651 and MCP4641 should be

```
final DigitalPotentiometer b = new MCP4651PotentiometerComponent(
                 bus, false, false, false, MicrochipPotentiometerChannel.B, MCP465
```

not

```
final MCP4651PotentiometerComponent b = new MCP4651PotentiometerComponent(
    bus, false, false, false, MicrochipPotentiometerChannel.B, MCP465
```

Additionally I could use factory-classes to build my instances of DigitalPotentiometer. I this situation I could replace my MCP4651 by any other digital potentiometer (and its implementation) and would have to change my code only at one place: the factory-class.

Another benefit of loose coupling is the possibility of writing tests which only test the class itself, not any depending classes. Mockito helps a lot for non-interface-depences but whenever you need to mock a class it is a sign of strong coupling.

So having this in mind, let's go back to the question of proper naming:

The usage should be

```
 final DigitalPotentiometer poti = new MCP4651PotentiometerComponent(...);
```

I have chosen "MCP4651" because everyone writing a software for a circuit using the MCP4651 knows that this is a potentiometer. But on the other hand, the person who wrote the code for creating the object-instance might be different to the person using this instance. So if someone needs to use the concrete class MCP4651 instead of DigitalPotentiometer it would be nice to have the word Potentiometer in the classname. This would make the software more readable. An example:

```
 public void increaseLedBrightnessToMaximum(
      final MCP4651PotentiometerComponent potentiometerUsedByPwmCircuit) { ...
```

and the loosing coupling-version:

```
 public void increaseLedBrightnessToMaximum(
      final DigitalPotentiometer potentiometerUsedByPwmCircuit) { ...
```

What also clears out after writing this example: The name of the class is not as important as the name of the parameter. The parameter's name is used within the code and is important for making the code understandable. For example:

```
    final int maxValue = potentiometerUsedByPwmCircuit.getMaxValue();
    potentiometerUsedByPwmCircuit.setValue(maxValue);
```

So the most classnames of my implementation are used inside of classes not of interest to the user and only affect the implementation's understandability not the user's code. So I would try to choose meaningful names especially for those classes visible to the user. On the one hand the user is the guy who wants to use the implementation and does not concern about the implementation (as long at it works ;-) I) and on the other hand it is someone like you and me who maintains the implementation -> two different kind of users.

Two rules I propose:

For public classes like MicrochipPotentiometerChannel: The name is OK since this is used inside of the user's factory methods and it is specific to those microship-potentiometers to provide multiple channels.

For "hidden" classes like MicrochipPotentiometerDeviceControllerChannel: I would use the old name in this situation. This class is only used inside of classes not visible to the common user. So there is no improvement to add "MicrochipPotentiometer" to the name since everyone reading/maintaining the code knows that this code is about microchip potentiometers. I think it is sufficient to provide this information in package-names for "hidden classes". Those classes will become more readable/maintainable doing so.

The last word on naming:

```
  MCP4651PotentiometerComponent
```

That fact that a MCP4651 potentiometer is a component is not important for the user. This is an internal classification of Pi4j and brings no benefit to the user if it is inside the classname. I think it is sufficient to find this information in the package-name. You also did not use DigitalPotentiometerComponent although this is the same situation. I always try to avoid use technical classifications for naming of classes or packages or parameters - always use names regarding the content. But you might say that this is niggling ;-). You might also say that I do not obey my own rules and I agree: It is hard keeping the user's view in mind on writing the implementation. Something instrumental for this situation is to do test-driven development: Think about the usage prior to the implementation.

But whatever you do I'm happy. The software works and I'm free to use it. What I told above are thoughts about all this - no more no less ;-)

Stephan

savageautomate:
OK, more refactoring changes ...   Let me know your thoughts ..

1.)  I moved all the MCP***\* classed under the "com.pi4j.component.potentiometer.impl.microchip" namespace.  This way all microchip related stuff is under one package.

2.)  I renamed all "MCP***_PotentiometerComponent" classes back to just "MCP**_*"   I agree that users consuming these will always know they are working with a potentiometer and the fact that it is a component isn't really important. 

3.)  I renamed the "DigitalPotentiometer" interface to just "Potentiometer".  I was not sure that adding the "Digital" prefix really adds much value since it will always be a digital potentiometer if you are programming something.  

4.)  I renamed the private classes back to "DeviceControllerChannel", "DeviceControllerDeviceStatus" and "DeviceControllerTerminalConfiguration".  I had actually noticed that on the first refactoring and meant to change them back, but must have forgotten.

5.)  I did swap the concrete class to the "Potentiometer" interface in the  test-main-classes in MCP4651 and MCP4641 ... however, these impls were accessing other methods that are not defined in the generic  "Potentiometer" interface.   So I added a new "MicrochipPotentiometer" that includes all the public methods from "MicrochipPotentiometerBase".  I don't think this really adds any benefit to the end user, but I added it anyway.  This change also resulted in needing to move "NonVolatileMode" out from inside the "MicrochipPotentiometerBase" class to public interface "MicrochipPotentiometerNonVolatileMode".

Changeset is here:
https://github.com/Pi4J/pi4j/commit/80ac4fd1c2e7078c78c2ac2b22ff5ced7e49c426

Again, I'm very open to any additional changes so we can nail this down for the release.

Thanks, Robert

RasPelikan:
Looks good. 

The only thing left is about which classes are located in which packages. As a user I would appreciate if all "internal" classes are hidden. So if I want to use a potentiometer, I would take a deeper look into the package where I found the component-interface.
At the moment I will find an impl-package and a microchip-subpackage including a bunch of interfaces and classes, also those classes which are not of my interest. OK, I will recognize the MCP4xxx classes and will start to use them.

So the only thing I would change is the package structure:
com.pi4j.component.potentiometer:
    Potentiometer -> OK. This is top-level. The interface has no dependencies
com.pi4j.component.potentiometer.microchip:
    MicrochipPotentiometer
    MicrochipPotentiometerChannel (needed by MicrochipPotentiometer)
    MicrochipPotentiometerNonVolatileMode (needed by MicrochipPotentiometer)
    MicrochipPotentiometerDeviceStatus (needed by MicrochipPotentiometer)
    MicrochipPotentiometerTerminalConfiguration (needed by MicrochipPotentiometer)
    MCP4xxx (they only need classes/enums/interfaces of the same package and
          MicrochipPotentiometerBase which is definitely an implementation-class)
com.pi4j.component.potentiometer.microchip.impl:
    anything else

In addition to this we could move the definition that MCP4xxx-classes are MicrochipPotentiometers into the classes itself and remove them from MicrochipPotentiometerBase. The MicrochipPotentiometerBase-classes itself does not need the interface definition. The only reason from being there is reducing duplicate code. But for the sake of transparency I would move the "implements MicrochipPotentiometer" into the MCP4xxx classes.

These things done the packages should be clear and if the user finds the MCP4xxx-class he needs he will also see that this class implements the Potentiometer-interface and the MicrochipPotentiometer-interface if he needs microchip-specific features.

What to do think about this?

savageautomate:
I do prefer your suggestions for the packages. 

OK, after moving the classes around to the new packages, this affects some of the impl code due to class visibility.  

I changed "MicrochipPotentiometerDeviceStatus" to an interface and then added "MicrochipPotentiometerDeviceStatusImpl" under the internal "impl" namespace allowing me to leave the the constructor private.

Do you think we should do the same for "MicrochipPotentiometerTerminalConfiguration" -- create an interface in the public namespace and create an impl under the "impl" namespace?

This still leave an issue inside "MicrochipPotentiometerChannel" where "com.pi4j.component.potentiometer.microchip.impl.DeviceControllerChannel" is now not accessible from inside the enum.

Changeset:
https://github.com/Pi4J/pi4j/commit/26618e3f3fef8b6aaa0f4d09faed1c936634607f

Thoughts on best way to handle?
Thanks, Robert

RasPelikan:
> Do you think we should do the same for "MicrochipPotentiometerTerminalConfiguration"
> -- create an interface in the public namespace and create an impl under the "impl" namespace?

No, because there is the method MicrochipPotentiometer#setTerminalConfiguration(...) which takes a MicrochipPotentiometerTerminalConfiguration-instance. So the user must be allowed to construct those instances (yes the user could reuse objects returned by getTerminalConfiguration, but this would take another I2C-call even if the user is not interested in the current configuration). You even might add the setters and a default-constructor so the class can be used by persistence-frameworks (hibernate, xstream) to serialize it as a persisted configuration.
The MicrochipPotentiometerDeviceStatus must not be constructed by the user, so splitting into an interface and an impl-class is the best solution.

> This still leave an issue inside "MicrochipPotentiometerChannel" where 
> "com.pi4j.component.potentiometer.microchip.impl.DeviceControllerChannel" is now 
> not accessible from inside the enum.

In MicrochipPotentiometerChannel remove the instance-variable
    private DeviceControllerChannel deviceControllerChannel;
and all references within the enum.

Afterwards add this method to DeviceControllerChannel:
    static DeviceControllerChannel valueOf(final Channel channel) {

```
    if (channel == null) {
        return null;
    }

    for (final DeviceControllerChannel dcChannel : values()) {

        if (dcChannel.name().equals(channel.name())) {
            return dcChannel;
        }

    }

    throw new RuntimeException(
            "There is no pendant for the given "
            + channel
            + "'! Maybe there was another channel introduced for new "
            + "devices but only to MicrochipPotentiometerChannel but "
            + "in DeviceControllerChannel it is still missing.");

}
```

Finally replace all occurrences of 
    channel.getDeviceControllerChannel()
(will be marked as errors by your IDE since you has removed this method) by
    DeviceControllerChannel.valueOf(channel)

Now everything should be fine :-)

When you are ready I will check it out any try the code with my ICs. I will be able to do this on Wednesday because now I have to stop working and in the next days I'm on a journey.

Stephan

savageautomate:
OK, the code is all checked in and a latest 1.0-SNAPSHOT deployed.  Thanks for all your help.    No problem on waiting until later in the week to get back to this.  I'll probably try to release 1.0 next weekend if all is going well.   

Safe Travels, Robert

RasPelikan:
Hi Robert,

I synced my fork to the HEAD of Pi4J/pi4j/develop. Having this done I rebuilt the library and could rerun the test (MCP4651#main(...)). Everything works fine :-)

Stephan

savageautomate:
Thanks for the update!

PR #140, opened 2015-03-16
New concurrency in I2CDeviceImpl and I2CBusImpl

RasPelikan:
Hello Robert,

as I wrote in the Pi4J Google group (https://groups.google.com/forum/#!topic/pi4j/aWzcyzrFEHs) I want to refactor I2CDeviceImpl to allow write/reading to multiple busses in parallel without blocking each other. Now I’m ready and sum up what’s new:

I2CFactory/I2CFactoryProvider/I2CFactoryProviderRaspberryPi/I2CFactoryProviderBananaPi:

The getInstance-methods throw an UnsupportedBusNumberException if the underlying system does not know the required bus-number. This is the only situation (of my changes) which effected customer’s code. Maybe you wish to remove this. At the moment, if a system knows about a bus but it isn’t configured it throws an IOException. This might be sufficient for unsupported bus-numbers, too. It’s up to you :-)

I added a method getInstance(int, long, TimeUnit) which takes a locking-timeout for all actions on the bus required. Yes this is new: locking is not up the device it is up to the bus!

I2CBusImplBananaPi/I2CBusImplRaspberryPi:

I changed class hierarchy so I2CBusImpl now is an abstract class and requires a specific class for using it. I2CBusImplBananaPi and I2CBusImplRaspberryPi are those classes.

I2CBus/I2CBusImpl:

As mentioned above now I2CBusImpl is an abstract class. I moved all read- and write-methods into this class but they are not part of I2CBus. They are part of the contract between I2CBusImpl and I2CDeviceImpl - maybe it is better to change the visibility to 'package' (now they are 'public'). I added a method ‚runActionOnExclusivLockedBus‘‚ which is used by I2CDeviceImpl for proper locking. This method may also be used by custom code to do several actions exclusively.

Writing and reading to com.pi4j.jni.I2C is now up to I2CBusImpl. I think it applies better to the real-world because a software writes to or reads from the bus to communicate with a device. Additionally now every shell (I2CDeviceImpl and I2CBusImpl beneath) enriches the users request for reading or writing: I2CDeviceImpl adds the address and error-handling; I2CBusImpl adds the filedesciptor and locking.

Additionally I added some code for proper handling of closing the bus.

I2CDeviceImpl:

Changed internals to match changes of I2CBusImpl. There are some inner-classes implementing the I2CRunnable required by I2CBus#runActionOnExclusivLockedBus. Those classes also may have been anonymous classes but I decided to name them for better readability.

Tests:

I added tests for I2CBusImpl and I2CDeviceImpl. For mocking of I2C I had to add powermock as a dependency and changed pom.xml to add a new JVM-parameter for unit-tests. Unfortunately there is a linker-problem reported on running the tests. This has to be ignored but it cannot be suppressed :-( -> I added a message which should calm the user :-)

I run the tests on a dual-core computer so threading should be save. I also run my MCP4651-main()-code which works as before. In the next days I should receive my Raspberry Pi 2. As it has four cores real concurrency happens. So I will retest it with the new hardware. But since the tests run already in a multithreaded environment I expect no surprises.

A last word on I2CBusImplTest: Testing concurrency uses Thread.sleep(…) to fake different situations. If the computer running the tests is busy by any other task those tests might fail. If the build is the only task the computer has to do everything works fine. I’m not happy about this but at the moment I have no idea how to improve this.

Stephan


RasPelikan:
Hi Robert,

I'm very sorry, but there is a bug: The method I2CBusImpl#getBus(...) takes an instance candidate which is used if there is no bus for the required bus-number available yet. This technique is used to hide all constructors entirely. I added this code recently and did not see that the method I2C#i2cOpen is called by the constructor. To clean this up move the lines of I2C#i2cOpen into a private method "open()" which has to be called by I2CBusImpl#getBus(...) if the instance candidate is used.

Additionally the timeout-parameters of the method I2CBusImpl#getBus(...) can be removed because the instance candidate is of a class derived from I2CBusImpl and so those values can be fetched by accessing the timeout-properties of the instance candidate directly.

Sorry for pushing buggy code,
Stephan

savageautomate:
Hi Stephan,

Is the bug fix you mention required for the changes in your pull-request -- or are these fixes necessary in the current merged codebase (see master branch)?  

Over the last weekend, in preparation for a 1.0 release, I merged the current release-candidate codebase from 'develop' into the 'master' branch.  Version 1.0 will be released from the 'master' branch.  This code is currently deployed as a '1.0-RC' snapshot.

I also went ahead and merged some serial API enhancements (v 1.1-SNAPSHOT) from a temporary placeholder branch into 'develop'.  

Thanks, Robert

RasPelikan:
Hi Robert,

the Fix is for this pull-request. I'm afraid to tell you that I don't know how to update the pull-request by fixing the bug in my own.

Stephan

savageautomate:
Hi Stephan,  sorry for the delay, I'll try to get this reviewed, merged, and a new 1.1-SNAPSHOT build out this week/weekend.  
Thanks, Robert

RasPelikan:
No problem. A few days ago I received my Raspberry Pi 2 and yesterday I managed to switch the second I2C bus to common-mode. 

Per default the bus 0 is in EEPROM mode in which common devices are ignored. As mentioned at http://www.raspberrypi.org/forums/viewtopic.php?f=44&t=102130 three things are necessary:
1. Add "dtparam=i2c_vc=on" to /boot/config.text which enables bus 0
2. Add "bcm2708.vc_i2c_override=1" to /boot/cmdline.txt which enables common-mode (= non-EEPROM-mode)
3. Upgrade the firmware to remove a bug preventing bus 0 to work. As I got no firmware update for my rasbian (I stick to 3.18.7) I hade to use a workaround for 3. which is to add
    echo 28 > /sys/class/gpio/export
    echo 29 > /sys/class/gpio/export
    echo in >  /sys/class/gpio/gpio28/direction
    echo in > /sys/class/gpio/gpio29/direction
to /etc/rc.local. Doing so my devices are listed by "i2cdetect -y 0". Maybe you like to place these hints at your Pi2-pinout page (http://pi4j.com/pins/model-2b-rev1.html)?

So today I will create a simple program using both busses at the same time. I will let you know about the outcome but I do not expect any surprises.

Stephan

RasPelikan:
OK, I wrote the simple program and everything works fine. I did not add this program to the pi4j-project because it uses the MCP4651 and the MCP4641 and so it is not easy to run for anyone else but me. If you like I could invite you on dropbox to see a video I made to prove it but this needs your email-address. Or you will wait until I could update my blog ;-)

albahrani:
I'm a bit concerned that the implementation of the I2C which is already "complicated" because of the locking and native access will get even more complicated by this pull request.

I really like the introduction of a "generic" approach concerning the amount of bus instances available on a certain raspberry model.

I understand the issue, but I think a solution shall be realized on a deeper layer, like wiringPi or he driver implementation. With any solution provided in pi4j you can only solve the problem of concurrency for a codebase running within a single VM. Multiple pi4j based VMs running on the same raspberry are still not covered and so are concurrency situations with the use of other libs and other programming languages.

My personal opinion is that adding this much code for such little benefit on the long run will bring pi4j in an unmaintainable state.

I would suggest removing the static locking in DeviceImpl which is the root of all evil here. It blocks all device instances ignoring that they may run on different bus instances.

RasPelikan:
My modifications address different aspects and for each aspect I concidered a lot of things:
- aspect "refactoring of raspberry pi / banana pi bus aquisition": The was a lot of duplicity, so I changed I2CBusImpl to be abstract and centralize any code needed by both implementations. This was also necessary to ensure that there is a singleton per bus which is mandatory for proper locking.
- aspect "locking to avoid corruption of single reads, writes or single write-reads": locking is good because accessing the bus by different communications may disturb each other. but it is up to the bus do locking because the bus should be locked and not the i2c-device. So removing the "static" from the current implementation will mean that locking is only done within one certain i2c-device but it is much more likely that accessing multiple devices will corrupt communication. Additionally the current implementation does not ensure that there is only one I2CDeviceImpl-object for each device so it is easy to disable that non-static device-locking by aquiring two devices of the same i2c-address. So locking has to be up the bus and not to the device!
- aspect "locking for long-running communications": It might happen that communication is bi-drectional. An example: You ask the device for reading data and it returns the number of bytes to read. The next communication is to read the bytes en block or byte-per-byte. At the moment it is not possible to block the bus for such kind of communication. May first solution was very close to achieve also this goal, so I also enabled this aspect by using Runnable and the single read/write/write-read uses this mechanism in the same way. But as far as I know this might not be necessary as long as single reads or writes are locked properly since this kind of locking is more up the communciation to a specific device that up to the entire bus.
- aspect "multiple jvms": I agree with you that it would be better to find a solution that also works using multiple jvms. In my opinion it is unlikely that multiple jvms will be used becauses jvm-startup takes 4 seconds on a Raspberry Pi and 1 second on a Pi 2. But from the point of view of providiing a library this restriction is not OK. On the other hand: This restriction way made long time before I added my change and I did not put this into question.
- aspect "WiringPi": I did a closer look at the current implementation before I started to change anything. As I could see, the native I2C-class does not use WiringPi to access I2C. It simply uses the file descriptors /dev/i2c-0 and /dev/i2c-1. The WiringPi-implementation seems to do the same way (https://github.com/WiringPi/WiringPi/blob/master/wiringPi/wiringPiI2C.c). So it seems to be a good idea to used https://github.com/Pi4J/pi4j/blob/master/pi4j-native/src/main/native/com_pi4j_wiringpi_I2C.c instead of https://github.com/Pi4J/pi4j/blob/master/pi4j-native/src/main/native/com_pi4j_jni_I2C.c. Does anybody know why there are two implementations for one thing?
- aspect "locking the bus": There is a possibility to lock the bus on a kernel-level: https://www.kernel.org/doc/htmldocs/device-drivers/API-i2c-lock-adapter.html. This may be used by any implemenation but must not. So if anyone (as WiringPi does) does not use this locking mechanism we have not made any step forward. I agree with you that this would be better but it seems to be out of my scope. I don't even could figure out whether accessing the way the com_pi4j_jni_I2C.c implementation does (by using the file desciptors /dev/i2c-*) would use this locking mechanism provided by the kernel. But in any way: This would not change anything on the way the java-code is modified by my pull-request. If simple would replace the java-locking mechanism by the kernel version. I could do so, but it has to be tested well in which way further access is blocked an whether this effects the systems load (this might be treated as io-wait and threrefore reduce the systems responsiveness, but I don't think so).

Nevertheless I need to access two busses simultaneously, so I need a useful solution. I could remove those Runnables and do locking within the I2CBusImpl on a per-read and per-write base. What do you think? Robert, any opinion?

albahrani:
Thanks a lot for taking my concerns serious. I really appreciate your fast and detailed answer.
You got me completely with the need for solutions to the different aspects you mention.

I'm just wondering if wrapping every Device read/write into an runnable and passing it to the executor is the best thing to do. I really would prefer a solution that has the as is "simple calls" and provides a kind of transaction wrapping around it.

I just did a quick implementation that should solve most of your aspects, except the wiringPi and driver related ones. It is really not tested. I could provide it as another Pull Request? Or is there a better way?

RasPelikan:
It seems you did not fork Pi4j on github. If doing so I could take a look there. I'm curious on your suggested solution! Do you have any public space where you can place a zip?

Wrapping into a "transaction" is to use a Runnable. This a common java-pattern. Constructing an instance of (for example) I2CDeviceImpl#WriteByteRunnable is nearly the same work-load as doing a "new Integer(123)". In the most cases a byte-array has to be constructed holding the data to be written. This is also of a similar complexity.
The most expensive instruction is the synchronization in I2CBusImpl which cannot be avoided if locking is done within the JVM. Constructing another object-instance (the runnable) is not that bad.

albahrani:
You are right. My fault. I thought you would map the Runnable to a queue.
Now there is a fork available. ;-)

RasPelikan:
I reviewed your changes:

I2CFactoryProviderBase: Your solution is quite simple and effective. At this circumstances your solution is more readable than mine. I do development for a long time and I tend to use design-patterns. Sometimes a pattern looks like an overkill but it is better to follow the pattern because the most patterns cover more things than visible at the very first moment. So the pattern I used is that a factory should not influence the behavior of the object it returns. It simple chooses between available implementations. So I provide two bus-implementations each returned by its provider. Only the implementation itself knows about its specific properties and behavior. At the moment the only difference between raspberry and banana pi is the number of busses available, so this is the reason why my solution may appear as an overkill. But maybe there will be any other "Pi" which uses different filenames then your solution will be to short.
But I think you did the right way to make busses reusable by opening and closing.

I2CBusImpl:
In the open- and close-methods you do not use try/finally for locking. This may cause a permanent lock if exceptions are thrown.
I think using a separate "closed" property is duplicate/redundant data. Using "fd == -1" within your closed-method is save. Especially if "I2C.i2cClose(fd);" throws an exception your "closed" property will run out of sync.
In your read- and write-methods: You first lock the bus and then call "checkIfBusIsClosed". If checkIfBusIsClosed throws and ioexception (as it is intended to do) the lock is permanent. One possibility is to move checkIfBusIsClosed into the try/finally block or to test on this condition before locking. A golden rule is: No code between locking and the try/finally-block. Sometimes it is better to move the lock-command into the try-block.
Any Exception (exactly Throwable) has to be catched within a finally-block because an exception thrown by the finally-block will hide any exception thrown by the try-block. Trust me: You don't want to search an exception never reported!
You add I2CRunnable to offer locking for series of actions but,
- you don't use it on your own
- you assume that those actions concern a certain device, but it is also likely that locking may be used for moving data from one i2c-device to another in a fast way. If locking the bus is done to get more speed and no interruptions your I2CRunnable is insufficient. I do not make any assumption for I2CRunnable expect that it may return a value.
  Your "executeTransaction"-method suggests that you do any kind of transaction. If a user of I2CBus does not take a look inside your code this may lead to wrong conclusion about its functionality.

What I miss is lock-timeout handling. If you run into a dead-lock the only way to get out of it is by stopping the entire JVM. Usually this is not acceptable because you are loosing your current state. Debugging this situation is also not easy. You will have to analyze thread-dumps but in most cases you need the current values of objects and properties to know what's going on. Using timeouts avoids most conflicts and gives you hints for improvements if a timeout occurs.

And I miss tests: You have some bugs (listed above) which could be avoided by writing proper tests. I know that tests are not a recipe for bug-prevention, but they ensure a minimum level of quality. I also found typos and wrong javadocs/comments (copy-paste-errors?).

I spent a lot of time to deliver a good quality and even than I had some bugs. Your code seems to be less modifications to achieve nearly the same functionality, this is done by making assumption I was not willing to do. Especially if you provide functionality to other people you have to think of any possible usage.  All my considerations about locking, possible usage, code-structuring and finally quality and testing led to the solution I sent in this pull-request. The parts being more complex are hidden to common users and are only of interest to the Pi4J supervisors. I do not think that your code (even if all bugs has been removed) is more suitable than mine is. The only aspect I would accept is doing locking at kernel-level but even that goal behind cannot be reached by this action.

I think I made my points and I think the decision is up to Robert.

RasPelikan:
@savageautomate : I added my fixes on my own (commits 6613cd8 and 3ab562f)

albahrani:
Thanks a lot RasPelikan for your feedback even if I think the last part is a bit harsh. My solution was just made to provide better access to my thoughts. I already mentioned before that it was a quick one and it is not tested at all.

savageautomate:
I've asked Daniel to see if he can review this merge request.  The original I2C implementation was contributed by him so it may make the most sense for him to take a look at it.  Sorry it's taken so long to get back to this request but we are trying to get caught up and address requests and bug fixes for the upcoming v1.1 release.   Thanks, Robert

RasPelikan:
Hi Robert,
it's funny: about 1 hour ago I visited the Pi4J homepage first since spring (as every year I do a lot of home and garden improvement in summer and electronics has to wait). You might have felt this :-).
Concerning my pull request: It would be fine to close this chapter.
In spring I started to add an implementation for a Microchip ADC series. I saw another guy added a component for ADCs in general: I will merge and test my work before doing another pull request.
I also did an implementation of a 1wire temperature-sensor but not in Pi4J as you posted in a blog that you don't see it as a part of Pi4J. Today I saw in the commit history that you withdrawed your decision and merged a pull-request from another guy. I will see if I can add any detail based on my implementation, but it is so simple that I do not expect anything. So to sum up: you can expect further pull-requests this fall/winter.
By the way: I also used your Stepper implementation and it worked fine as everything at Pi4J!
Stephan

savageautomate:
Hi Stephan,

Glad to hear from you. Yes W1 did make its way in .. by popular demand LOL
:-)
There has been some recent activity in the ADC/DAC impls.  You might check
and see if the MCP chip you are targeting already exists in Pi4J's GPIO
extension library.  But we would always welcome more :-)

Trying to get things caught up in the project for a 1.1 release, maybe
towards the end of the year.

Just remember (which I'm sure you already do) to base your pull-requests
off the 'develop' branch.

Thanks, Robert

natdan:
Where was I when all this here started? (Actually I know, but that's different story). I, again, apologise from being completely disconnected. I'll pull it in on my local machine and go through your argument here (there's quite a lot to digest) and try to add my view to it. In general, I see things started moving in positive direction  - even if we go only with what @RasPelikan said. I do get the @albahrani 's concerns, but I am not 100% sure what is better: ignoring synchronisation and let it (currently) not being handled in lower layer or implementing it even if we do not that it is not quite correct (if given multiple VMs and other clients).
In general, my intuition is to try to provide as clean interface as possible, and then not prevent multiple different (or even very similar!) implementations that try to solve different or even same problems in different ways. So, watch this space. Hopefully I'll be quicker than with one wire interface (and there's quite lot of work there and it is going to be changed a lot, I am sure) - at least my first even stud/partition wall is nearly finished and it seems to be true. And it wasn't simple straight wall O: )

natdan:
OK, I have scanned through the code and here are some comments:
1. Tabs vs Spaces - it turned it be less readable because of mixing both.
2. I don't believe that runActionOnExclusivLockedBus should be on the interface
3. That mechanism (runActionOnExclusivLockedBus + Runnable) seems a bit of a overkill. If you have 3-4 I2C communications per frame ~ we'll have many new objects created per second on slow system with limited memory. GC overhead added to all of it. I don't mind it ended up being one of different implementations to play with, but somehow it is a bit of overkill for 'ordinary' use
4. I really like idea of read/write methods being on I2CBusImpl (not 100% sure if we should elevate them on the interface as well. Probably not). 
5. Making I2CBusImpl abstract makes perfect sense. 
6. UnsupportedBusNumberException is great idea and it should go in separate file.
7. Use of RuntimeException is not what I would do - in many places we can do with checked or unchecked existing exceptions (like NullPointer exception instead of RuntimeException in runActionOnExclusivLockedBus).
So, in general I like the refactoring but I wouldn't go that far with creating new object for each 'action' and then delegating it to the bus... 
I'll look more into it...

RasPelikan:
Hi Daniel,
concerning "another object per communication": I did not had this in my mind. But do you really think this object is that expensive? I'm not sure which applications might cause so much communication. I have ADCs, Digipots, an 3-axis gyroskop and an NFC n-tag. The only IC which might be used heavily is the ADC. Oh I forgot I also used a PIC's ADC and I2C to fetch the values. I could fetch 5000 samples per second. But as far as I know SPI matches better needs of intensive communication. What I want to say is that I think that I2C is an expensive way of communication and I think it will not be used for high frequent communication. But I might be wrong.
However, I agree to not waste resources on restricted systems. So what do you suggest?
Stephan

natdan:
I wouldn't just throw away that implementation as wrong, but also wouldn't want to dictate it as the way we must go with. So, leave it in 'experimental' implementation - an extension of I2CBusImpl for instance, but remove it from I2CBus interface. Simple standard implementation should just direct all I2C reads and writes to I2CBus (as per your suggestion) and there we can use simple synchronise (or alternative you proposed) to serialise access to the particular i2c bus (as you originally suggested). Even synchronisation is quite slow operation, but you might be right saying that we shouldn't allow two threads spoil integrity of a bus...
And why overcomplicating things where we can very little return if any. 

BTW I get your reasoning what i2c device should and shouldn't be and I agree with it.

RasPelikan:
Hi Daniel,

yesterday I did some refactorings which should make things more readable. Let me explain:
- 575af7c: I replaced I2CRunnable by the standard interface Callable<T> which matches better. I did not know about Callable in spring.
- 1340f96: Now the wrapping of IOException is not necessary any more, so I removed it.
- 6d4bb8c: I combined all checks made for every I2C operation into a separate method used by each operation for less code.
- 4a98686: I moved all classes based on the Callable interface to an extra class (I2CDeviceImplHelper) which cleans up I2CDeviceImpl and improved readability.
- fcc8ba0, fcc8ba0: I removed duplicate code.

Afterwards I was not happy at all. So further I did:
- e10249f: Moved all static classes based on the Callable interface from I2CDeviceImplHelper back to I2CDeviceImpl as anonymous classes and deleted I2CDeviceImplHelper. This has as advantage that I could directly access variables of the calling scope so this removed a lot of boilerplate code. The disadvantage is that error-handling appears in every anonymous class and is nearly the same. But I think this is OK since the code is more readable. I think the code is now fairly easy to read.

I also did some minor changes and merge your current develop-branch so the pull request becomes mergeable again.

I also would like to comment your last comment at least to let you know that I though about everything you wrote:

"I wouldn't just throw away that implementation as wrong, but also wouldn't want to dictate it as the way we must go with. So, leave it in 'experimental' implementation - an extension of I2CBusImpl for instance, but remove it from I2CBus interface."

I thought about this and I think moving my implementation to an alternative class produces a lot of additional code and I think you want me to make things more easier. I also want to bring out that I wrote a two excessive junit-tests (95,7% coverage for I2CDeviceImpl; for I2CBusImpl coverage could not be calculated by the eclipse-plugin I don't know why, but coverage should be at least 85%) to prove that this code is not experimental.

"Simple standard implementation should just direct all I2C reads and writes to I2CBus (as per your suggestion) "

yes, as it is

"and there we can use simple synchronise (or alternative you proposed) to serialise access to the particular i2c bus (as you originally suggested). Even synchronisation is quite slow operation, but you might be right saying that we shouldn't allow two threads spoil integrity of a bus..."

I did not introduce ReentrantLock (see https://github.com/Pi4J/pi4j/blob/master/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CDeviceImpl.java). I just restructured existing code into the method "runActionOnExclusivLockedBus" for minimizing lines of code by improved functionality.
What I introduced is the timeouts used for locking to avoid "uncontrolled locking-situation". I don't think that this will happen but I'm sure you also did ever try to find a deadlock in a non-debugging environment what is not that funny (digging thread-dumps). Using timeouts prevents users to restart their apps in such situations and Exceptions thrown by timeouts can easily be logged or catched. This is little code for a better world :-)
Replacing by normal "synchronize" would be OK but would be less functionality that in the current master: you used reentrant locks with "fair" usage. simple synchronization does not honor the order of lock-aquisition. I even think that using synchronization would safe only a couple of lines. Additionally unlocking becomes implicitly by closing the synchronize-block which is less transparent. But if you prefer it I can remove ReentrantLock.

"And why overcomplicating things where we can very little return if any."

I'm sure it is now not that complicated as I removed a lot of boilerplate code. The pattern of using Callable (previously I2CRunnable) is a common way of delegating work into another context. It is even so common that java 8 introduced a language-feature for this: lamda expressions. Just for fun I made another branch in which I used lambdas instead: https://github.com/RasPelikan/pi4j/blob/lambda/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CDeviceImpl.java. If you take a look at the delta (https://github.com/RasPelikan/pi4j/compare/develop...RasPelikan:lambda) you will see that it is nearly the same, just a couple of lines are gone. In a java-8 environment I would prefer lambdas.

I hope you will be happy with my refactorings and am looking forward to your response.

Stephan

natdan:
Huh. There's quite a lot to digest. Please give me some time + ability to respond in code if I need to! :)
Also, I like you gave some thinking to it and put even more effort in testing!

RasPelikan:
Hi Daniel, could you take a look?

natdan:
Certainly. I have spared at least one day for pi4j and other RPi development over Christmas break.

natdan:
OK. Finally got some time. In order for this to be done 'properly' I've merged all your changes (and sorted as much as I could merging conflicts) in new branch called bus-reorganisation. Until we get final, stable version that can be perfect playground for all bus changes. 

I've yet to go through all changes you've done, but I've got some time over last week to read all you put there and I agree with most of it. Also, I have couple of ideas how what you were after might be done, too. I'll write very soon again.

natdan:
I've done more work in the branch where I have merged your changes. Please do inspect it. I am planning to make it even more streamlined (i2c implementation) - return back to completely non-synchronised version in base I2CBus implementation and allow configuration which would be added at top of it which will implement it they way you suggested. That way 'core', simple, fast, non-dependent on extra VM features to be used with due care, or platform (or our providers) to be asked to provide fully synchonised implementation of the bus. Further, I think synchronised version of the bus should really be just an adapter over I2CBus interface allowing synchronisation over ANY implementation of the bus itself (no matter if we use bus as file based, kernel implemented implementation, bit banging one, some bus extenders - even as bridges over different other buses like serial to i2c, etc). That we will keep all you suggested but at the same time allow porting pi4j to VMs not capable of such synchronisation like LeJOS and/or HaikuVM. What do you think of this suggestion?

The moment I get another couple of hours of work I'll submit code to the branch and ask you to check it. Also, I have submitted issue for Java 8 and if we even go that way I'll replace Callable interface with lambdas when methods called.

natdan:
Oh, forgot to say that one of the biggest re-factoring I've done is to move all synchronisation to the bus implementation itself as I don't believe it is I2CDevice's concerns. Bus is the place which needs to organise what is sent over and how. 
BTW speaking of different synchronisation solutions - there's one where we will have a thread per bus waiting in the background for something being put in a FIFO queue and process bus communications as they (commands/ r/w actions) put the queue. Not sure we should implement it as well - maybe just as an experiment - and let users decide what kind of bus management to use.

RasPelikan:
Thank you for refactoring. I reviewed the changes you made. I agree with your version. I also plugged my breadboard using two I2C-devices (digital potentiometers: MCP4641 and MCP4651) on different busses of my Raspberry Pi 2 and run a stress test (in addition to the existing unit-tests). It works pretty well: running 10,000 commands (setting random pot-values immediately one after one) on one I2C-device takes 4.946 seconds (user: 1.79, sys: 0.49). Running 20,000 commands in parallel (two threads using two cores and two different I2C busses) takes 4.96 seconds (user: 2.09, sys: 0.770). As you can see, it consumes 0.014 seconds for additional 0.58 seconds (user and sys). It seems that the two busses do not block each other :-)

savageautomate:
closing this pull-request as the code changes for this pull request are now refactored in temporary branch: https://github.com/Pi4J/pi4j/tree/bus-reorganisation 

That brach is pending a final merge into `develop`

savageautomate:
@natdan 

Are the proposed changes in https://github.com/Pi4J/pi4j/tree/bus-reorganisation ready to be merged? 

Thanks, Robert

natdan:
@savageautomate Not yet. I'll manually cherry-pick fixes and changes that do not alter break any existing interface. I'll do it tonight.

PR #145, opened 2015-03-30
Address thread leaks around use of the debouncer

ryangardner:
It can be difficult to know if there is a thread leak or if things are behaving normally when the threads use the default "pool-N-thread-Y" names.

When looking at a thread dump from my application, it wasn't clear if there was a thread leak in pi4j or just a ton of generic threads - this change helps identify that the threads came from pi4j and what they are used for

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/concurrent/DefaultExecutorServiceFactory.java
+++ b/pi4j-core/src/main/java/com/pi4j/concurrent/DefaultExecutorServiceFactory.java
@@ -34,69 +34,87 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class DefaultExecutorServiceFactory implements ExecutorServiceFactory {
-    
+
     public static int MAX_THREADS_IN_POOL = 25;
-    private static List<ExecutorService> singleThreadExecutorServices  = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


ryangardner:
I updated the pull request - I didn't rip out the "newSingleThreadExecutorService" method - but I don't see any uses of it. It seems like it would be a good idea to remove that since keeping the list of those created single thread executors around so they can be shutdown later I think is the primary source of the leak.

I added a GpioEventExecutorService and refactored the existing wrapper that was just around a ScheduledExecutorService to be against any ExecutorService and then extended it to have a version that adds delegate methods for the ScheduledExecutorService to disable the shutdown methods on it. 

I built a version locally with these changes and am running it now. My device should be triggering some IO events in the next few hours and I'll check the thread dumps later to make sure it's not leaking threads anymore.

ryangardner:
I'm pretty sure this fixes the thread leak issue - I just saw the first several GPIO event on my system and saw a single thread in the cached executor pool be created. A minute later (the default timeout for a CachedThreadPool) the thread is no longer in my thread dumps.

savageautomate:
I think leaving the deprecated 'newSingleThreadExecutorService()' method is fine for 1.0 and we can remove it completely in 1.1.  

Question .. should some caller request a newSingleThreadExecutorService(), should we continue to add it to the internal managed list so that the shutdown can be called later? 

```
        // add new instance to managed collection
        singleThreadExecutorServices.add(singleThreadExecutorService);
```

I realize this is one of those questions like ... _"If a tree falls in the forrest does it make a sound?"_  ... since we never invoke it internally.  Should some other user code have gotten a reference to the GpioFactory.getExecutorServiceFactory() and used it to get their own singleThreadExecutorServices()  instance for some reason ...  (albeit, this is probably highly unlikely)

PR #157, opened 2015-04-29
Updated several example files

martinbeentjes:
I updated several example files. Most of them contained infinite for-loops with a Thread.sleep(500); in them. I think a while(true) is a better solution and it will not require any Exception handling.


savageautomate:
I implemented most suggestions in the example files.  I did not implement the change where thread sleeps were omitted in favor of a continuous while(true) loop.  While I do agree it looks simpler, I think it really spikes the CPU usage.     Thanks and sorry for taking so long!

martinbeentjes:
Well, that is a good thing to notice. Thanks for your response, I will keep this in mind in the future. :-)

Issue #32, opened 2013-02-09
Add 1-wire support using GPIO pin


savageautomate:
Feature request to add support in Pi4J for the GPIO bit-banging 1-wire Linux kernel driver: 

pi@raspberrypi:~$ sudo modprobe wire
pi@raspberrypi:~$ sudo modprobe w1-gpio
pi@raspberrypi:~$ sudo modprobe w1-therm 

See this article for details:  http://webshed.org/wiki/RaspberryPI_DS1820

marcandreuf:
I implemented an example in JAVA using the modprobe driver and reading from the file. Just like the article. At least I could read temperature values in a Raspberry pi :-)

https://github.com/marcandreuf/sunfounder-sensors-raspi-4j/blob/master/src/main/java/org/mandfer/sunfunpi4j/Ex16_Ds18b20.java

I hope it helps.

ademcaliskan:
I would suggest an static function for listing all  available wired devices as an Array or List.

Adem

DataWorm:
Schups code looks fine, but it is a few months old. Until now there has been no reaction on his merge request... Is there a chance, that it will get merged soon or got this issue postponed for a specific reason?

natdan:
I've only got active now (someone was of impression that he's following this project while he wasn't). I'll go through the code over weekend and do the necessary. Sorry for the delay...

savageautomate:
BTW, it will get merged into the 'develop' branch which means that it will be available in the 1.1-SNAPSHOT builds.

DataWorm:
Seems to take some more weekends? ;-)
Could we help in some way? Is there something that needs to be improved (apart from some javadoc comments) before it could be merged?

natdan:
All help appreciated. Do you know how to lay laminate? Or at least remove existing tiles from kitchen floor? I think I got confused by Robert's message thinking he has already done it. Will start looking into it tonight. Sorry about it.

DataWorm:
I think I could learn it on-the-fly. But free support for such tasks is limited to Berlin! ;-)

natdan:
OK. I've skimmed through Schups's code and I am not 100% sure what to do. Code itself is fine and I'll pull b76d065 and f16c54c - it will, at least buy us some time. Only thing is that it, probably, will be just WIP or initial (working) starting point for One Wire interface. 
[...]

PR #162, opened 2015-05-26
#32 Initial implementation for 1 Wire based devices using /sys/bus/w1

schup:
I've started implementing support for the 1 Wire protocol using the kernel drivers.

This is not ready to be merged but I wanted to get a view if it makes sense to implement it this way.

The Master is part of the pi4j-core and uses SPI (Service Provider Interface) to load individual device implementations.

An implementation for the DS18B20 temperatur sensor is in the pi4j-device.

See pi4j-example/src/main/java/W1TempExample.java for example usage.

Comments welcome :-)


natdan:
OK. I have merged it. Also, I had to put @Ignore tags on two tests (hashCode and equals methods test) in TmpDS18B20Test.java as I didn't know why they were failing. Could you, please, look into it and see if those tests are still valid and that there's an error in the code or tests do not stand anymore?

BTW an observation - not an policy or enforced thing - just something I adopted a while ago: In Eclipse you can add actions on save which would remove trailing (and only trailing) white space from lines.

PR #163, opened 2015-05-28
Fixes clear methods of LCDBase.java

neyold:
clear() was only clearing 2 row LCDs, this fix solves the problem. Method that
clears specific row, starting from a specific column for specific length had a major
bug. This commit fixes that too.


neyold:
You're absolutely correct!

PR #188, opened 2015-10-16
initialize test class state for every @Test

jjfraney:
Use @Before instead of @BeforeClass to initialize fixture for
each test.  This eliminates the possibility of a dependency between
@Test methods.


savageautomate:
Thanks for the contribution

PR #192, opened 2015-11-15
Added rounding mode to fix Arithmetic Exception

cjm27:
The updateResolution method invokes an Arithmetic Exception:

```
Exception in thread "main" java.lang.ArithmeticException: Non-terminating decimal expansion; no exact representable decimal result.
```

This should fix the error. But I am not sure, since I have not update the .jar dependency file yet. I have some projects using Raspberry Pi and GPIO in mind, and I would love to be able to program in Java. This is my first pull request to this project, and I will continue to contribute. Thanks!


natdan:
I've merged it. It makes sense - even if it doesn't fix the problem.

savageautomate:
Thanks for the contribution!

Issue #183, opened 2015-10-09
MCP23008GpioProvider. - java.lang.ArrayIndexOutOfBoundsException: 128


GeorgeProost:
Hi, Thanks for the  great wrapper.

I have located a possible addressing issue in MCP23008GpioProvider or thereabouts. Cannot seem to address all the gpio pins on the MCP23008 chip. GPIO_07 (ie 8th one).

I 'think' it has to do with the gpio addressing of 128 for pin07

Any ideas?

Exception in thread "main" 
java.lang.ArrayIndexOutOfBoundsException: 128
    at com.pi4j.io.gpio.GpioProviderBase.getPinCache(GpioProviderBase.java:69)
    at com.pi4j.io.gpio.GpioProviderBase.export(GpioProviderBase.java:98)
at com.pi4j.gpio.extension.mcp.MCP23008GpioProvider.export(MCP23008GpioProvider.java:127)
    at com.pi4j.io.gpio.GpioProviderBase.export(GpioProviderBase.java:79)
    at com.pi4j.io.gpio.impl.GpioPinImpl.export(GpioPinImpl.java:158)
at com.pi4j.io.gpio.impl.GpioControllerImpl.provisionPin(GpioControllerImpl.java:517)
    at com.pi4j.io.gpio.impl.GpioControllerImpl.provisionDigitalOutputPin(GpioControllerImpl.java:669)
    at sfcountertester.SFCounterTester.<init>(SFCounterTester.java:52)
    at sftestsuite.SFTester.main(SFTester.java:71)

Regards 

George

GeorgeProost:
Hi,

I think the issue lies deeper, not in pi4j ...

Regards

George

GeorgeProost:
Hi, I am reopening the issue, does seem to have to do with pin caching.
If there is any more info required, I can try to get it to you. I include the source:
package sfcountertester;

import com.pi4j.gpio.extension.mcp.MCP23008GpioProvider;
import com.pi4j.gpio.extension.mcp.MCP23008Pin;
import com.pi4j.io.gpio.GpioController;
import com.pi4j.io.gpio.GpioFactory;
import com.pi4j.io.gpio.GpioPinDigitalOutput;
import com.pi4j.io.gpio.PinState;
import com.pi4j.io.i2c.I2CBus;
import java.io.IOException;

/**
 *
- @author gproost
  */
  public class SFCounterTester implements Runnable {
  
  private int numRelays;
[...]

PR #200, opened 2015-12-20
Issue 183

RasPelikan:
Fixed for MCP23017. If the solution is in the way you like it, please apply it to the other gpio-providers in this package (MCP23008 und MCP23S17).

I could not test every situation since my circuit only uses it for output-mode. So input-mode and interrupt-handling is not yet tested. The changes I made are not so sophisticated. I hope I find all situations and adopted them properly. If you review them I think it will be fine (four eyes principle).

Stephan


savageautomate:
I do like your approach (_retrospectively_), but the real underlying issue with #183 was a bug introduced while attempting to optimize the pin cache, this has been fixed and should be working properly in the latest builds.  Thanks!

PR #201, opened 2015-12-21
Create EasyDriver

cjm27:
A simple class for use with Easy Driver Stepper Motor Driver.

### DIFF
--- a/pi4j-device/src/main/java/com/pi4j/component/motor/EasyDriver
+++ b/pi4j-device/src/main/java/com/pi4j/component/motor/EasyDriver
@@ -1,0 +1,201 @@
+package com.pi4j.component.motor;
+
+import com.pi4j.io.gpio.*;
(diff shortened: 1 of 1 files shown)


natdan:
Nice! :) 

Did you maybe consider defining an interface for driving stepper motors?

cjm27:
Thanks!
There is a StepperMotor interface already but does not seem to be compatible with the chip, nor many other drivers. I will look into more designs of stepper motor drivers and suggest a better interface for it.
I will also need to write better constructor(s) for this class. It currently needs every pin to be defined, while one may only need a few of the pins for it to work.

savageautomate:
Manually merged into 'develop' branch. 
https://github.com/Pi4J/pi4j/commit/0e4ac3d12cf1e434a16a38953d2185cd04e72eaf

PR #204, opened 2016-01-04
Update MCP23S17GpioProvider.java

Balakosh:
Supports now all possible Addresses of the MCP23S17 at the A0 A1 and A2 pins.


savageautomate:
I manually copied your changes into the '_develop_' branch where the current 1.1-SNAPSHOT builds get build from.   

Thanks for the contribution!

PR #205, opened 2016-01-04
Adding detection of GPIO Pin Edges

Brainsucker92:
Detects the Edges of GPIO Pins and triggers an Event.


savageautomate:
REF: https://github.com/Pi4J/pi4j/commit/8850bbb6407dd5bb643d11fb7c5efe4b39301a94

Merged the MCP pin changes.
Implemented the getEdge() logic in the existing pin event rather than creating a new event, listener and detector.
Removed pulse events, listener, detector; pulses can be more complex than the current impl allowed for.  Pulses may not always start with signal HIGH and end on signal LOW.

PR #209, opened 2016-01-11
GpioInterrupt modified to use a java thread for polling. Fixes #207

### LINKED ISSUE #207: Interrupt-based GPIO is dead slow
I'm trying to use Pi4J com.pi4j.wiringpi.GpioInterrupt to read GPIO pin for events at 2kHz (clock ticks), but unfortunately I'm only getting callbacks at the rate of 40Hz, i.e. very slowly. 

Reading the code, my quick guess is that it's the AttachCurrentThread/DetachCurrentThread pair in pi4j-native/src/main/native/com_pi4j_wiringpi_GpioInterrupt.c that is causing the slowness. It also causes each callback call to look like it's coming from a different thread (new name and id).
- Is there a reason why the thread is Detached after each callback?
- Is there a reason the poll loop is not written in Java with just the call to poll being native code. This way the Thread would be a normal Java thread, no need for Attach/Detach and the whole native code would be a one-liner?

hvesalai:
Conflicts:
    pi4j-core/src/main/java/com/pi4j/wiringpi/GpioInterrupt.java
    pi4j-native/src/main/native/com_pi4j_wiringpi_GpioInterrupt.c
    pi4j-native/src/main/native/com_pi4j_wiringpi_GpioInterrupt.h

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/wiringpi/GpioInterrupt.java
+++ b/pi4j-core/src/main/java/com/pi4j/wiringpi/GpioInterrupt.java
@@ -56,11 +56,12 @@ import java.util.Vector;
  * @see <a href="http://www.pi4j.com/">http://www.pi4j.com/</a>
  * @author Robert Savage (<a
  *         href="http://www.savagehomeautomation.com">http://www.savagehomeautomation.com</a>)
+ * @author Heikki Vesalainen
  */
 public class GpioInterrupt {
 
     private static Vector<GpioInterruptListener> listeners = new Vector<>();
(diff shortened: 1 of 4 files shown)


savageautomate:
Whoops, did not mean to close this.  It may have auto-closed when I collapsed and removed the 'develop' branch when v1.1 was released.  With v1.1 released now, I am ready to move forward on adding this change to 1.2-SNAPSHOT.  

@hvesalai would you mind re-submitting this PR on the master branch? 

Thanks, Robert

PR #212, opened 2016-01-19
Make i2c polling on the MCP23017 IO expander dynamic

GeVanCo:
I've experienced that polling the MCP23017 IO expander on a 50ms base is not really needed.  Therefore, I propose to make the polling time
dynamic, with a default of 50ms.  This way,  people who don't want to change this, can use the SW the way they used it before.  Others can put the polling time to a value of their preference.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBus.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBus.java
@@ -39,9 +39,17 @@ import java.io.IOException;
  */
 
 public interface I2CBus {
-
-    public static final int BUS_0 = 0;
-    public static final int BUS_1 = 1;
+    
+    public static final int BUS_0  = 0;
+    public static final int BUS_1  = 1;
+    public static final int BUS_10 = 10;
+    public static final int BUS_11 = 11;
+    public static final int BUS_12 = 12;
+    public static final int BUS_13 = 13;
+    public static final int BUS_14 = 14;
+    public static final int BUS_15 = 15;
+    public static final int BUS_16 = 16;
+    public static final int BUS_17 = 17;
 
     /**
      * Returns i2c device.
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CBusImpl.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CBusImpl.java
@@ -52,7 +52,31 @@ public class I2CBusImpl implements I2CBus {
 
     /** Singleton instance of bus 1 */
     private static I2CBus bus1 = null;
-    
+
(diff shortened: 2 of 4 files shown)


savageautomate:
Can you base your pull request on the 'develop' branch?  All new changes get committed/merged into the 'develop' branch and are included in 1.1-SNAPSHOT builds.   Thanks, Robert

mosulache:
Hi,
I've introduced this modification and after using it with a setting of 1 second (DEFAULT_POLLING_TIME = 1000) I discovered a bug.

Normally I see no reason why one would not choose 1 second checking. In my case, I check for a water level sensor, so 1 second is more than enough. What is happening is that there is an easy way to miss the sensor change! I think the bug is in the monitor:

 int pinInterruptB = device.read(REGISTER_INTF_B);

```
                        // validate that there is at least one interrupt active on port B
                        if (pinInterruptB > 0) {
                            // read the current pin states on port B
                            int pinInterruptState = device.read(REGISTER_GPIO_B);

                            // loop over the available pins on port B
                            for (Pin pin : MCP23017Pin.ALL_B_PINS) {
                                int pinAddressB = pin.getAddress() - GPIO_B_OFFSET;

                                // is there an interrupt flag on this pin?
                                //if ((pinInterruptB & pinAddressB) > 0) {
                                    // System.out.println("INTERRUPT ON PIN [" + pin.getName() + "]");
                                    evaluatePinForChangeB(pin, pinInterruptState);
                                //}
                            }
                        }
```

I think the problem is here:
 int pinInterruptB = device.read(REGISTER_INTF_B);

This registry seems to be "active" (signaling the event) for a limited period of time, and with a larger polling time it is easy to skip reading the event. I tested and I was able to reproduce with 2 sensors (inputs) and pin status readings at 1000.

To fix this, I just commented this registry check, and then go directly into checking the status of each pin.

Maybe someone can tell me why there are 2 levels of checks? As far as I can see, if no event happens, there is still a check (involving communication with MCP) and if there is an event there are 2 calls  ( the second one to get the actual pin values). Then why not checking all the time full pin status : a single registry!

Namely the second reading only:
  int pinInterruptState = device.read(REGISTER_GPIO_B);

---

I have several hardware implementations that use several devices on i2c bus  and all of them are featuring some problems with MCP23017 (chip is blocked after a while - from 1 day to 30 days), and as such I tried to figure out race conditions / misusage of the pi4j lib or faults in the pi4j itself.

By the way, dynamic i2c polling is an useful idea. 
Maybe this is not the best place to place my comment, but it concerns a bug that is mostly revealed by this change.

GeVanCo:
@savageautomate :  I've moved my changes to the develop branch.  I hope all necessary steps are taken to achieve what you wanted (I'm quite new in the GitHub wow, hence...).

PR #239, opened 2016-05-30
unlock if an IOException occurs

francescou:
bugfix: the lock will never be unlocked if an IOException is thrown


natdan:
Quite right. Thanks!

PR #245, opened 2016-06-12
Add support for MCP3424 ADC converter.

falkena:
Hi,
I've added support for MCP3424 Delta-Sigma ADC converter.
Can be used with MCP3422/MCP3423 also.


savageautomate:
Looks good, thanks for the contribution!

falkena:
You are welcome.

Issue #260, opened 2016-08-03
Native I2C Impl should return real errno's


colinrgodsey:
Currently the impl only returns a -10001, -20001 or -30001. The negative return value from the i2c system calls will only return -1 in the 'less than 0' case. This is just to indicate that you should check errno for the real error code. 

Working on a PR that will return the negative errno from the native impls (still with the same offset for ioctl/read/write). These errnos will be used on the java side to form meaningful IOExceptions as a subclass.

colinrgodsey:
addressed in #289

Issue #259, opened 2016-08-03
Native I2C Impl can't possibly be thread safe


colinrgodsey:
I understand there's locking around bus operations, but there's not across multiple busses and there's a shared static buffer that will be used for the entire process, and could cause issues under parallel access:

https://github.com/Pi4J/pi4j/blob/release/1.1/pi4j-native/src/main/native/com_pi4j_jni_I2C.c#L44

colinrgodsey:
Fix im considering for this:

~~using pthread_key_create/pthread_setspecific/pthread_getspecific~~

EDIT: actually moved all operations to direct access, or stack-allocated

PR #261, opened 2016-08-03
i2c cleanup

colinrgodsey:
deferred to PR #264

addresses #260 #259 
- reduced ioctl calls
- removed shared static buffer in native impl
- direct referencing of stack-allocated data and java byte array contents
- native calls now return errno
- i2c impl throws an IOException subclass with error code
- bounds checks for buffers

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CIOException.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CIOException.java
@@ -1,0 +1,54 @@
+package com.pi4j.io.i2c;
+
+import com.pi4j.jni.I2C;
+
+import java.io.IOException;
+
+public class I2CIOException extends IOException {
+    int rawCode;
+    String baseMessage;
+
+    /**
+     * @param message Exception message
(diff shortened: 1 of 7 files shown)


natdan:
I might be slow to pull it and compile it on RPi as I am quite away from one - it is a silly season here (read: holidays). Will do it as soon as I can.

colinrgodsey:
@natdan sounds good 👍

pi4j-core/src/main/java/com/pi4j/io/i2c/I2CIOException.java
```diff
@@ -0,0 +1,54 @@
...
+        this.rawCode = Math.abs(rawCode);
+    }
+
+    /**
+     * Gets the POSIX code associated with this IO error
+     *
+     * @return POSIX error code
+     */
+    public int getCode() {
+        return rawCode - getType() * 10000;
```
colinrgodsey, line 27: my lazy ass forgot what a modulo is it seems

colinrgodsey:
deferred to #264

PR #265, opened 2016-08-07
Alternate I2C impl

colinrgodsey:
This is an amendment to my other PR. This removes the JNI r/w calls for i2c and just leaves the ioctl slave select function. We can use Java's **RandomAccessFile** for simple read/write operations (and state handling), and can use just the ioctl JNI call for slave select. 

This _does_ break API compatibility as far as exceptions go. The exceptions thrown now are mostly just the standard java IO exceptions you would see when working with sockets or files. With this, I decided this should be its own PR that may be considered for future versions/releases if exception message compatibility matters (they're still IOExceptions, but the message is totally different).


natdan:
I really like it! Shame I cannot give it a go yet, but I think we should seriously consider it. Worst case scenario (or intermediate step) we can keep same interface and your impl. Will return back to you as soon as I get closer to RPi! :)

colinrgodsey:
All good! The extra time let me mull it over more ;)

savageautomate:
I'll be happy to do a build and push the SNAPSHOT artifacts out as soon as @natdan approves and merges the PR.  The I2C stuff is sort of his domain :-)

colinrgodsey:
@natdan so ive got another insane prospective patch....

I made a new class, LinuxFile, which extends RandomAccessFile and gives us ioctl methods directly right there. We can directly ioctl via a long, a byte array, or a flat 'structured' byte array. Main reason for doing this is I finally found a device that needed stop-less transactions. This method of using ioctl would allow any advanced user to implement literally any other I2C/smbus related ioctl (or pretty much anything that uses ioctl), including custom transactions. 

For the structured data, you can manually assemble all structures needed in a single byte array, then there's an additional array that lists indexes (from the flat byte array) where pointers are, and provides the offset (inside the byte array) that it points to. This is allows us to use pointers, as long as we point to data that's inside the given byte array:

byte array:
byte[] {0xFF, 0xFF, 0, 0, xx, xx, xx, xx, 1, 2, 3, 4, 5}

offset list (int array) even size only ([index of pointer, index where pointer points to]):
int[] {
   4 // index of pointer marked by xx, 
    8 // where it points to
}

It works pretty flawlessly! Biggest gotchas are native word size (provided via a static on LinuxFile), and C struct padding. The padding is annoying, but anyone use to GCC struct padding/alignment can figure it out (basically pointers always start on a native word boundary).

Both arrays can also be re-used, making this 0-alloc if given enough care.

colinrgodsey:
(full branch here: https://github.com/colinrgodsey/pi4j/tree/linuxfile)

heh, pardon my scala, but this is an example of what using the advanced ioctl looks like (works with 64b or 32b):

``` scala
      def i2cTransact(writeBuffer: Array[Byte], writeOffset: Int, writeSize: Int,
        readBuffer: Array[Byte], readOffset: Int, readSize: Int): Unit = {
      import LinuxFile.wordSize

      val maxLength = 16 + 16 * 2 + readSize + writeSize
      val numMessages = if (readBuffer != null) 2 else 1
      val numPointers = numMessages + 1

      val data = new Array[Byte](maxLength)
      val offsets = new Array[Int](numPointers * 2)
      val buf = ByteBuffer.wrap(data)

      var offsetIdx = 0

      def idx = buf.position

      def advance(n: Int) = buf.position(idx + n)

      def putPointer(): Int = {
        val curIdx = idx

        advance(wordSize)

        curIdx
      }

      def setPointer(ptrLoc: Int, dataOffset: Int) = {
        offsets(offsetIdx) = ptrLoc
        offsets(offsetIdx + 1) = dataOffset

        offsetIdx += 2
      }

      buf.order(ByteOrder.LITTLE_ENDIAN)

      //i2c_rdwr_ioctl_data   8 or 16 bytes
      val msgPointer = putPointer() //msgs
      buf putInt numMessages //num
      if(wordSize == 8) advance(4) //pad

      //i2c_msg #1 - write  12 or 16 bytes
      val messageOffset = idx
      buf putShort addr.toShort //addr
      buf putShort 0 //flags
      buf putShort writeSize.toShort //len
      advance(2) //pad
      val writeBufPtr = putPointer() //buf

      val readBufPtr = if (readBuffer != null) {
        //i2c_msg #2 - read
        buf putShort addr.toShort //addr
        buf putShort I2CConstants.I2C_M_RD.toShort //flags
        buf putShort readSize.toShort //len
        advance(2) //pad
        putPointer() //buf
      } else -1

      //append read/write buffers
      val writeBufOffset = idx
      buf.put(writeBuffer, writeOffset, writeSize)

      val readBufOffset = idx
      if(readBuffer != null) advance(readSize)

      setPointer(msgPointer, messageOffset)
      setPointer(writeBufPtr, writeBufOffset)
      if(readBuffer != null)
        setPointer(readBufPtr, readBufOffset)

      tryError(inst.ioctl(I2CConstants.I2C_RDWR, data, offsets))

      if(readBuffer != null)
        Array.copy(data, readBufOffset, readBuffer, readOffset, readSize)
    }
```

colinrgodsey:
hmm, so my LinuxFile impl has started to take on a mind of its own... it's pretty handy! Added support for direct ByteBuffers and non-truncating memory mapping (the NIO memory mapping tries to truncate, unfortunately).  I managed to implement a 0-alloc webcam streamer using just the new LinuxFile ioctls. Just needed to copy the v4l2 headers over to java consts and that was about it. I'm thinking it might be starting to scope-creep pretty heavily- not sure it's best for the pi4j library, but its definitely handy for anyone wanting to interface with linux device drivers without writing JNI.

natdan:
I'll be back onto integrating what you've done to pi4j now... and as for v4l stuff - that's fantastic. I think it should find it's way to pi4j - why not! If there isn't a area already where it fits we can easily invent one. Worst case scenario - another project in pi4j set of projects. I'll return back to you as soon as I sit seriously and pull all your changes and start playing with them! :)

colinrgodsey:
Great! The LinuxFile impl has definitely been crucial with interfacing with some drivers(playing with a /sys/mem version of gpio currently), and it's about as direct as you can get while avoiding the pitfalls and hassle of jni code. 

I like that it also borrows from the idea of user space drivers in general- the ioctl is supposed to allow direct device control from user space, were basically allowing userspace drivers in a managed  VM. Seema natural to extend it there, even thought it's not cross platform anumore.

colinrgodsey:
so I think I might try to clean up a branch and make a temporary PR so you can view the changes.  heh, oddly enough though, the last thing im going back-and-forth over is how to handle exceptions. It could be nice to expose the errno code in some ways, but to keep with the java-ness of things, it might make more sense to translate the errno code directly in to the standard error string. This is what java does internally, so we should be able to produce IOExceptions that have the same textual messages as the native ones

natdan:
I've given it a go and some rudimentary example I have works same as it was before. So, from my perspective there's absolutely no reason why LinuxFile wouldn't find its way to pi4j. And i2c begin re-written using it. Only thing I'll do as soon as this merges go over your Java code and re-format it to confirm to Java Coding Conventions (and add blocks after if statements). 

Also, soon it is in, we can progress with clean up and revisit need for locks - especially if they are anyway done on much lower level.

PR #289, opened 2016-11-14
LinuxFile impl - Alternative interface for ioctl devices

colinrgodsey:
This PR offer an extension of java's RandomAccessFile, which provides JNI methods for accessing: mmap/munmap, ioctl (structured et al), errno/strerr.

Should address: #262 and #260 

This implementation currently needs some cleanup, but is otherwise totally functional:
* Formatting
* Cleaner impl for mmap'd ByteBuffers. Should be able to explicitly munmap them, and should be able to be casually GC'd (like the NIO mmap BB impl, but with explicit unmapping supprt)
* mmap'd buffers should maybe close when the file descriptor is closed? Not sure. Most mmap drivers allow the mapped region to be open after the FD is closed, so not sure.
* Might need a cleaner impl with I2CBus and I2CDevice. I've been recently using LinuxFile directly in my project, so haven't *completely* verified the I2Cxxx classes (although I was using them for a bit just fine).
* Some of the comments are wrong, and the example in the comments for structured ioctl is definitely wrong. Could probably just make a wiki article for the 'how to' and link there.

Example I2C device impl using just LinuxFile (again, pardon my scala...):

EDIT: the provided sample references I2CConstants, which is the java-ized version of i2c-dev.h

https://gist.github.com/colinrgodsey/71bf701e33deeb7ebcf38a2b146e345f


savageautomate:
Do you feel that this change set is ready to be merged into master?

Thanks, Robert

natdan:
Hi Robert, I believe it is but I am struggling to find time to test it. Could you, please, give me another weekend or so.

savageautomate:
Sure thing .. no rush really. I was just trying to close out some of the open tickets.  :-)

colinrgodsey:
I'm also back on the radar now- going to try and produce a cleaner version of this at some point tonight. I think I finally figured out how to handle mmap'd ByteBuffers in a sorta sane way. 

The goal with the buffers is to allow them to be GC'd without thought, or to be explicitly unmapped via the LinuxFile interface. The issue with the explicit unmapping is being able to invalidate the existing ByteBuffer at that time- otherwise it hangs around with a pointer to some random chunk of memory- potential issues there for sure.

So, I'm thinking that upon explicitly unmapping, I can reset the pointer back to 0. This will cause the ByteBuffer to throw an exception (un-initialized exception, i think) upon access.

natdan:
This is now manually merged in. I'll close it for now.

savageautomate:
Please see issue: https://github.com/Pi4J/pi4j/issues/310

jvalkeal:
@colinrgodsey 

Right so my error was thrown from 
```
} catch (NoSuchFieldException e) {
    throw new InternalError();
```
It might have been good idea to pass exception instead of error out with empty `InternalError`. 

Not sure how this was supposed to work(think `cleaner` doesn't exist in `Buffer`) as generally accessing fields via reflection is very very bad idea. I do it in tests but never in a framework code.
```
public abstract class Buffer {

    /**
     * The characteristics of Spliterators that traverse and split elements
     * maintained in Buffers.
     */
    static final int SPLITERATOR_CHARACTERISTICS =
        Spliterator.SIZED | Spliterator.SUBSIZED | Spliterator.ORDERED;

    // Invariants: mark <= position <= limit <= capacity
    private int mark = -1;
    private int position = 0;
    private int limit;
    private int capacity;

    // Used only by direct buffers
    // NOTE: hoisted here for speed in JNI GetDirectBufferAddress
    long address;

    // Creates a new buffer with the given mark, position, limit, and capacity,
    // after checking invariants.
    //
    Buffer(int mark, int pos, int lim, int cap) {       // package-private
```

PR #293, opened 2016-11-30
Software SPI Bus implementation using a JNI

NewLunarFire:
Based on the proof-of-concept in this repo:https://github.com/TeamNeko/SoftSPI, thought I would offer to add it to Pi4J

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
@@ -23,7 +23,7 @@ import com.pi4j.wiringpi.Gpio;
  * this project can be found here:  http://www.pi4j.com/
  * **********************************************************************
  * %%
- * Copyright (C) 2012 - 2016 Pi4J
+ * Copyright (C) 2012 - 2017 Pi4J
  * %%
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as
--- a/pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
@@ -1,0 +1,182 @@
+package com.pi4j.io.spi.impl;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+import com.pi4j.io.spi.SpiDevice;
+import com.pi4j.io.spi.SpiMode;
+import com.pi4j.jni.SoftSPI;
+import com.pi4j.wiringpi.Gpio;
+
+/*
+ * #%L
+ * **********************************************************************
+ * ORGANIZATION  :  Pi4J
+ * PROJECT       :  Pi4J :: Java Library (Core)
+ * FILENAME      :  SoftSpiDeviceImpl.java
+ *
+ * This file is part of the Pi4J project. More information about
+ * this project can be found here:  http://www.pi4j.com/
+ * **********************************************************************
+ * %%
+ * Copyright (C) 2012 - 2016 Pi4J
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Lesser Public License for more details.
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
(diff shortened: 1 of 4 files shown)


pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
```diff
@@ -0,0 +1,182 @@
...
+ * **********************************************************************
+ * ORGANIZATION  :  Pi4J
+ * PROJECT       :  Pi4J :: Java Library (Core)
+ * FILENAME      :  SoftSpiDeviceImpl.java
+ *
+ * This file is part of the Pi4J project. More information about
+ * this project can be found here:  http://www.pi4j.com/
+ * **********************************************************************
+ * %%
+ * Copyright (C) 2012 - 2016 Pi4J
```
vborcea: It shhould be  Copyright (C) 2012 - 2017 Pi4J

pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * Creates a Software SPI Device. 
+	 * @param chipSelect Pin to use for Chip Select
+	 * @param clock Pin to use for Clock
+	 * @param masterOut Pin to use for Master Out Slave In
+	 * @param masterIn Pin to use for Master In Slave Out
+	 * @param speed Speed of the SPI Port
+	 */
+	public SoftSpiDeviceImpl(int chipSelect, int clock, int masterOut, int masterIn, long speed, SpiMode mode) {
+		super();
```
vborcea: super is called implicitly so it can be removed

pi4j-core/src/main/java/com/pi4j/jni/SoftSPI.java
```diff
@@ -0,0 +1,25 @@
+package com.pi4j.jni;
+
+/**
```
vborcea: New classes should have project license header

pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
NewLunarFire: No problem, but I actually submitted the PR in 2016.

vborcea (approved):

pi4j-core/src/main/java/com/pi4j/io/spi/impl/SoftSpiDeviceImpl.java
```diff
@@ -0,0 +1,182 @@
...
+	@Override
+	public ByteBuffer write(ByteBuffer data) throws IOException {
+		return ByteBuffer.wrap(write(data.array()));
+	}
+
+	@Override
+	public byte[] write(InputStream input) throws IOException {
+
+        // ensure bytes are available
+        if(input.available() <= 0){
```
iqnev, line 134: An 'if-else-if' statement without a terminating 'else' clause may allow execution to 'fall
through' silently. If the terminating else clause is intended to be unreachable code, it is advisable that it throws a RuntimeException to alert the user of an internal error

natdan: Else in this case is the rest of the function. This pattern is similar to:
```
if (condition) {
    return;
}
... something else
```

ii is equivalent to:

```
if (condition) {
    return;
} else {
    ... something else
}
```

pi4j-native/src/main/native/com_pi4j_jni_SoftSPI.c
```diff
@@ -0,0 +1,80 @@
...
+
+JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *jvm, void *reserved)
+{
+  return JNI_VERSION_1_2;
+}
+
+JNIEXPORT jbyteArray JNICALL Java_com_pi4j_jni_SoftSPI_readWrite(JNIEnv * env, jclass class, jint cs, jint clk, jint mosi, jint miso, jlong speed, jboolean cpol, jboolean cpha, jbyteArray data)
+{
+  int i;
+  long nanos = 500000000 / speed;
```
colinrgodsey, line 30: I would move this magic number here to a #define for the file

pi4j-native/src/main/native/com_pi4j_jni_SoftSPI.c
```diff
@@ -0,0 +1,80 @@
...
+  digitalWrite(cs, LOW);
+
+  for(j = 0; j < arraySize; j++)
+  {
+    in = 0;
+    out = outputData[j];
+
+    for(i = 0; i < 8; i++) {
+      if(!cpha)
+	digitalWrite(mosi, (out & 0x80) >> 7); //Write current bit
```
colinrgodsey, line 50: maybe also move the _(out & 0x80) >> 7_ operation to its own macro:

```c
#define HIGH_BIT(byte) ((byte & 0x80) >> 7) 
```

This is probably more of a personal preference

NewLunarFire:
I have copied most of this from the base SPI implementation in Pi4J. Perhaps it would be better to fix the issues in the base implementation, them have the software SPI inherit the class and override only the necessary functions.

FDelporte:
Sorry, we will not be adding additional features to pi4j v1, but are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

PR #306, opened 2017-02-09
String values are compares using "==", not equals

vbauer:


savageautomate (commented):
Any benefit of using Objects.equals() vs the String::equals() method ... like pin.getName().equals(getName())  ?

natdan:
Hi, nicely spotted but your suggestion is not the best. Instead of Object.equals we should really have "64".equals(System.getProperty(...) kind of code. Do you want to make another pull request, fix this one or leave this to me? :)

savageautomate:
@natdan,

@vbauer is using ```Objects.equals``` which I believe delegates to the "equal" method of the first argument.  (Don't feel bad ... I had to look it up :-) )

vbauer:
> Hi, nicely spotted but your suggestion is not the best. Instead of Object.equals we should really have "64".equals(System.getProperty(...) kind of code. Do you want to make another pull request, fix this one or leave this to me? :)

No problem, I've just updated this PR

> @vbauer is using Objects.equals which I believe delegates to the "equal" method of the first argument

Yes, it is just a null-safe way to check equality

natdan:
@savageautomate Yes, I you are right Object.equals can easily do the job. It is just suggestion to follow construct/pattern/best practice(?) used in Java the most. Object.equals is perfect if any of two arguments can be null. When we have one constant string then the easiest way is:

`"string".equals(object)`

We know "string" is never null and equals method of the string caters for nulls of argument. As I said - just pattern I've been dealing with last 20 years of coding in Java...

natdan:
@vbauer cool! 👍 

BTW just just reminded me of another ocd-alike change we can do:

Where it makes sense to statically import static methods. In above case Object.equals - importing equals is quite fine. Rule of thumb I've been using for a while regarding static imports is:

1. IDE nicely hides imports so we don't worry having more lines of code - they are not visible in everyday coding
2. Importing well known, frequently used methods is fine. Methods like Arrays.asList, org.junit.Asserts.assertXXX, in unit tests matchers like is(), any(), etc... 
3. With some other things - if methods is clear and descriptive: MyEntityObject.newMyEntityObject is good candidate as code would read x = newMyEntityObject()  (don't quote me on example - it is not the best)
4. And don't if method is way to generic and it would be hard to deduct. For instance for immutables generated method is ImmutableMyTransferObject.builder() importing just "builder()" means very little as we have no idea what type it is...

I've merged your changes now and when I get chance I'll look into this clean up, too. But feel free to give it a go if you suffer of same (OCD) as I do O: )

PR #309, opened 2017-02-10
Update Maven plugins

vbauer:
Minimum version of Maven was also configured in the "prerequisites" section
It it the minimal version which is required by all plugins

```
[INFO]
[INFO] Require Maven 3.0.5 to use the following plugin updates:
[INFO]   org.apache.felix:maven-bundle-plugin .......................... 3.2.0
[INFO]
[INFO]
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,11 @@
 		<module>pi4j-distribution</module>
 	</modules>
 
+    <!-- PROJECT PREREQUISITES -->
+	<prerequisites>
+		<maven>3.0.5</maven>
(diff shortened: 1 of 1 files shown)


savageautomate:
I'll have to pull this change down locally later tonight and test a full snapshot release build to make sure all the updated plugins continue to work for a full build.  I remember having some trouble a while back when upgrading to newer plugins.  Those issues may have already been resolved so hopefully a merge of this PR will be coming later tonight.  Thanks, Robert

vborcea (approved):

Issue #310, opened 2017-02-12
InternalError with LinuxFile


jvalkeal:
Ok, looks like something got broken in master possibly with https://github.com/Pi4J/pi4j/commit/65e541bc393c45db1108b60c0154729bbcc5ef0e#diff-e64dac8dc75e58e99bd691f019538a6b or something else with related changes with a `LinuxFile` because I'm now getting
```
java.lang.InternalError: null
	at com.pi4j.io.file.LinuxFile.<clinit>(LinuxFile.java:87) ~[pi4j-core-1.2-SNAPSHOT.jar!/:na]
	at com.pi4j.io.i2c.impl.I2CBusImpl.open(I2CBusImpl.java:123) ~[pi4j-core-1.2-SNAPSHOT.jar!/:na]
	at com.pi4j.io.i2c.impl.I2CProviderImpl.getBus(I2CProviderImpl.java:48) ~[pi4j-core-1.2-SNAPSHOT.jar!/:na]
	at com.pi4j.io.i2c.I2CFactory.getInstance(I2CFactory.java:76) ~[pi4j-core-1.2-SNAPSHOT.jar!/:na]
	at com.pi4j.component.lcd.impl.I2CLcdDisplay.<init>(I2CLcdDisplay.java:131) ~[pi4j-device-1.2-SNAPSHOT.jar!/:na]
```

Just a head up as I didn't have time to investigate further. This `LinuxFile` looks to be a new stuff.

colinrgodsey:
@jvalkeal what JVM and version are you using? The calls that are failing are reflection calls to an internal API, which i fear might not exist in the version you're using.

jvalkeal:
```
$ java -version
java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) Client VM (build 25.65-b01, mixed mode)
```
This is the arm version but code is same for all later versions afaik.

jvalkeal:
I quickly went through jdk7/jdk8/jdk9 sources and none have that `cleaner` field in `java.nio.Buffer`.

colinrgodsey:
@jvalkeal you know, you're absolutely right about that one. im very confused now how I got this build running for test at all. its starting to look like i leaked some untested commit into that last release branch. will get a fix branch up soon

colinrgodsey:
[...]

PR #314, opened 2017-02-15
Fix LinuxFile reflection init (#310)

colinrgodsey:
**Addresses #310, related to new LinuxFile implementation merged in with #289**

Fixes incorrect java reflection field lookup in the LinuxFile static initialization. 

Not sure how this snuck past testing, but I was sure to test the fail case this time, along with the fix.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/file/LinuxFile.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/file/LinuxFile.java
@@ -75,7 +75,7 @@ public class LinuxFile extends RandomAccessFile {
 
             addressField = Buffer.class.getDeclaredField("address");
             capacityField = Buffer.class.getDeclaredField("capacity");
-            cleanerField = Buffer.class.getDeclaredField("cleaner");
+            cleanerField = dbb.getDeclaredField("cleaner");
             directByteBufferConstructor = dbb.getDeclaredConstructor(
                     new Class[] { int.class, long.class, FileDescriptor.class, Runnable.class });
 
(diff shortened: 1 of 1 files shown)


jvalkeal:
I'll try this later this week(work life getting in a way).

Btw what's a motivation to use reflection as it's something you should not do against java framework libs. Most tools flags these as errors. Also I believe most of  `sun.misc.*` stuff is going to get removed with jdk9 http://openjdk.java.net/jeps/260

colinrgodsey:
So the cleaner specific stuff is for there mostly for mmap, for video support. Unfortunately FileChannel doesn't support mmap on files with undefined size (tries to make them larger), which doesn't work for the V4L2 devices. 

Long story short, we're having to hack the DirectByteBuffer a bit... which most likely isn't a maintainable option, but should work for the moment.  There should hopefully be a better option in the future, but worst case scenario if the apis go missing, we throw a friendly error and not just pop a bunch of InternalErrors. 

Overall, the memory-mapped buffer implementation is still being thought over... I will definitely take java 9 and a reflection-less solution into consideration. Portability would be nice. The mmap stuff should probably have been removed before the LinuxFile release anyways, given I still don't feel great about the impl.

colinrgodsey:
I'm going to jump the gun and merge this in now, as it is effectively bricking I2C on master.

savageautomate:
FYI, new 1.2-SNAPSHOT build now deployed that include these changes.

colinrgodsey:
👍  thanks @savageautomate

jvalkeal:
@colinrgodsey just tried latest snapshot, all good!

colinrgodsey:
great!

PR #316, opened 2017-02-20
Raspberry supports only 2 i2c busses

falkena:
Hi,

quick&dirty fix for number of available busses on RaspberryPi.

Better solution should be try to open an device on bus.
And if open failes throw an UnsupportedBusNumberException exception.

Issue #303, opened 2017-02-07
I2C BUS choices omit half the range of the PI's PCA9548 /dev/i2c-# range


TomOke:
With the PI 3's reduction of accessible I2C buses to just i2c-1 work was done to support a pca954x i2c multiplexor. Work was done by 6by9 with a description found through: 
https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=141517

With the addition of the dts overlay, and the presence of a pca954x chip (up to 8 channels with the 9548) one can then get:
/dev/i2c-1, i2c-3, i2c-4, i2c-5, i2c-6, i2c-7, i2c-8, i2c-9 and i2c-10, where the driver handles the Multiplexor's selection mask as buses are addressed (the mux is on the physical i2c-1 bus).

However the Raspberry Pi implementation:
 pi4j/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CFactoryProviderRaspberryPi.java

public class I2CFactoryProviderRaspberryPi extends I2CProviderImpl {

    @Override
    protected String getFilenameForBusnumber(int busNumber) throws UnsupportedBusNumberException {
        if ((busNumber < 0) || (busNumber > 3 && busNumber < 10) || (busNumber > 17)) {
            throw new UnsupportedBusNumberException();
        }

        return "/dev/i2c-" + busNumber;
    }
}

Directly takes the supplied BUS number and generates a /dev/i2c-# file name, but specifically only accepts the range 0-4, 10-17, which misses half the mux's range. The constants in:
 pi4j/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBus.java
mirror this limit, omitting 5, 6, 7, 8, 9.

Since the code will already correctly detect and report an error with a non-existant i2c dev path it should be safe to simply drop the " || (busNumber > 3 && busNumber < 10)" check and accept the full range. Adding the additional BUS_# constants would be nice, but since they directly devolve to a number one could get by without that.

natdan:
You are right. I think we should allow those bus numbers as well. Maybe even just remove restriction to everything bigger or equal to 0. I'll remove that restriction now from I2CFactoryProviderRaspberryPi and add missing bus numbers to I2CBus.java.

TomOke:
Thank you very much. 

Tom

natdan:
I've pushed changes to master. Check them out and if you are happy close the issue.

TomOke:
[...]

PR #317, opened 2017-02-21
Raspberry supports only 2 i2c busses 

falkena:
Reworked version of #316.
@TomOke Can you test this PR please? I can't due to lack of PCA954* chipsets
@natdan Any chance to test it on another (not only) fruit/vegetable boards?

Not sure, if
```javascript
public static I2CFactoryProvider getI2CFactoryProvider(Platform platform)
```
can be removed also


TomOke:
I will try to give it a shot, but it may be a number of days away.

pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CFactoryProviderRaspberryPi.java
```diff
@@ -35,7 +35,7 @@
 
     @Override
     protected String getFilenameForBusnumber(int busNumber) throws UnsupportedBusNumberException {
-        if (busNumber < 0) {
+        if ((busNumber < 0) || (busNumber > 1) {
```
iqnev: Missing  the last closing bracket in if clause

natdan:
I have a question: is there a specific reason why you are fetching all bus numbers at the creation of factory and keep it cashed and not just checking if particular file exists at the creation of the bus? If I understand they way kernel driver for PCA954 works those buses might appear after it is attached - after you have cached fact that there are only two buses - not whole bunch of them.

pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CFactoryProviderRaspberryPi.java
falkena: This file does not exist anymore in last commit.

falkena:
I2C is not hot-plug bus as specified. Means, device tree should be here, if Pi4J is spawned.
It exists mosfet circit for decoupling slaves and master. So you are right: If somebody
use this circit, it can happen, that devices "comes" and "goes avay" any time. Actually, this
problem is not only about PCA, but any I2C IC on the bus.

This and performance was a reason to cache all "/dev/i2c-*" entries found on first start.
I can fetch "/dev/i2c-*" entries on every getBus - call for sure.

colinrgodsey:
I've got a PCA954* board in-hands. Going to try and get it running today and should be able to test this PR. MacroFab screwed up my Pi header placement ( :1st_place_medal: ) and I don't get a replacement until tomorrow, but I should be able to rig it up today (or if I give up, tomorrow when I get the header).

pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CProviderImpl.java
```diff
@@ -30,25 +30,42 @@
...
+                String[] tokens = path.split("-");
+                if(tokens.length == 2) {
+                    devices.put(Integer.valueOf(tokens[1]), path);
+                }
+            }
+        } catch (IOException exception) {
+        }
+    }
+
     public I2CBus getBus(final int busNumber, final long lockAquireTimeout, final TimeUnit lockAquireTimeoutUnit) throws UnsupportedBusNumberException, IOException {
```
colinrgodsey: Would it make sense to just add an additional overload for getBus that takes a bus path string (in addition to the bus-number caching above)?

I know in my project, I configure everything using the actual bus path itself (it makes sense when dealing with multiple devices/busses that register under /sys), and have to do a reverse mapping anyways.

pi4j-core/src/main/java/com/pi4j/platform/Platform.java
```diff
@@ -148,28 +145,7 @@ public I2CFactoryProvider getI2CFactoryProvider() {
     }
 
     public static I2CFactoryProvider getI2CFactoryProvider(Platform platform) {
-        // return the I2C provider based on the provided platform
-        switch(platform) {
```
colinrgodsey, line 152: 👍  I like the reduction here

pi4j-core/src/main/java/com/pi4j/io/i2c/impl/I2CProviderImpl.java
falkena: May be it's generally better to get information over  /sys/bus/i2c/devices instead of /dev
In this case, overload will be ok for my opinion. For io operations we should use /dev
Should work at least on Raspberry. Not sure about banana and so on...
@natdan @savageautomate Any reason not to implement overload?
Is I2CFactory.java better for this?

http://unix.stackexchange.com/questions/176215/difference-between-dev-and-sys

colinrgodsey: Generally the paths in /dev are used for 'exported' non-root accessible entry points, while /sys (sysfs) by default is generally root only. I think this also comes down to just common-practices in Linux kernel drivers, which of course can differ from the standard. 

I think the only thing you can count on is that i2c-dev requires a file path/handle of some sort to operate.

EDIT: I don't think the i2c /sys interface offers an i2c-dev file. Believe it has to be exported to /dev/ specifically. 

EDIT2: i2c-dev (the kernel module) is apparently the actual module responsible for exporting the user-accessible /dev/i2c-* files. So even the i2c device drivers themselves have no control over that, and everything should be there in /dev. If it's not there, then something went wrong with the bus device or driver.

falkena: Hacked down detection via /sys just now: Runs perfect on Raspberry/jessie as pi user.
Attention: pi is in gpio group.

Kernel version:
Linux 4.4.38-v7+ #938 SMP Thu Dec 15 15:22:21 GMT 2016 armv7l GNU/Linux

```
ls -la /sys/bus/i2c/devices
insgesamt 0
drwxr-xr-x 2 root root 0 Feb 23 17:22 .
drwxr-xr-x 4 root root 0 Jan  1  1970 ..
lrwxrwxrwx 1 root root 0 Feb 23 17:23 i2c-1 -> ../../../devices/platform/soc/3f804000.i2c/i2c-1
```
Patch file in attachment. Please rename from zip to patch
[sys_dev.zip](https://github.com/Pi4J/pi4j/files/797412/sys_dev.zip)

What system are you running?

colinrgodsey: Sorry, totally butchered my last response. 

Are you using the /sys path just for finding busses (and not the io)?

colinrgodsey: Alright, read through the i2c-dev module docs again. It says you should actually rely on /sys/class/i2c-dev/ for getting the /dev/ names. That folder should contain all the names for the i2c-dev devices (and only those values) available in /dev. 

I'm assuming it's similar as filtering through /dev/i2c-*, but using /sys/class/i2c-dev/ does tell you witch /dev/ devices are actually bound to the i2c-dev module. Suppose it rules out any chance of naming conflicts in /dev.

colinrgodsey: https://www.kernel.org/doc/Documentation/i2c/dev-interface

> Now, you have to decide which adapter you want to access. You should
> inspect /sys/class/i2c-dev/ or run "i2cdetect -l" to decide this.
> Adapter numbers are assigned somewhat dynamically, so you can not
> assume much about them. They can even change from one boot to the next.

```c
  int file;
  int adapter_nr = 2; /* probably dynamically determined */
  char filename[20];
  
  snprintf(filename, 19, "/dev/i2c-%d", adapter_nr);
  file = open(filename, O_RDWR);
  if (file < 0) {
    /* ERROR HANDLING; you can check errno to see what went wrong */
    exit(1);
  }
```

falkena: May be a solution is to follow symlinks:

```
ls -la /sys/bus/i2c/devices
insgesamt 0
drwxr-xr-x 2 root root 0 Feb 23 17:22 .
drwxr-xr-x 4 root root 0 Feb 23 19:46 ..
lrwxrwxrwx 1 root root 0 Feb 23 17:23 i2c-1 -> ../../../devices/platform/soc/3f804000.i2c/i2c-1

ls -la /sys/class/i2c-dev/
insgesamt 0
drwxr-xr-x  2 root root 0 Feb 23 20:54 .
drwxr-xr-x 49 root root 0 Jan  1  1970 ..
lrwxrwxrwx  1 root root 0 Feb 23 20:54 i2c-1 -> ../../devices/platform/soc/3f804000.i2c/i2c-1/i2c-dev/i2c-1

```
Can't evaluate this properly, since only one i2c bus available on my pi2.

falkena: Pushed my last changes. Would be great if you can test them with PCA.

natdan:
> This and performance was a reason to cache all "/dev/i2c-" entries found on first start.
I can fetch "/dev/i2c-" entries on every getBus - call for sure.

I don't think that's a good reason. Bus is normally obtained once and infrequently so there isn't much of a performance issue if we check if file exists at the time someone is to fetch it. But, on the other hand, if on the startup bus wasn't ready or ready yet, then you won't be able to access it at all. Or worse if it was but disappeared (and that happens with i2c devices in some cases) then we are back to square one. I don't believe caching buses at the start up of JVM (or first access to those classes) is correct approach. Rest of the code - removal of unnecessary tree of classes is spot on! :)

pi4j-core/src/main/java/com/pi4j/io/i2c/I2CFactory.java
```diff
@@ -78,6 +78,28 @@ public static I2CBus getInstance(int busNumber) throws UnsupportedBusNumberExcep
...
+     * The timeout for locking the bus for exclusive communication is set to DEFAULT_LOCKAQUIRE_TIMEOUT.
+     *
+     * @param sysfs The sysfs device path. E.g /sys/bus/i2c/devices/i2c-0
+     * @return Return a new I2CBus instance
+     * @throws UnsupportedBusNumberException If the given bus-number is not supported by the underlying system
+     * @throws IOException If communication to i2c-bus fails
+     * @see I2CFactory#DEFAULT_LOCKAQUIRE_TIMEOUT
+     * @see I2CFactory#DEFAULT_LOCKAQUIRE_TIMEOUT_UNITS
+     */
+    public static I2CBus getInstance(String sysfs) throws UnsupportedBusNumberException, IOException {
```
iqnev: If 

> sysfs
is `null` the current code will  throws nullpointer

natdan: Without going deeper into the code philosophically speaking - shouldn't code throw null pointer exception if an argument of that method is null? BTW I do understand that it would be better to have assertion of some kind and give more precise error ('argument sysfs is null' for instance) as NPE message.

natdan:
@falkena On Thursday/Friday (sorry that I can't do it sooner) I'll try to make another pull request similar to what you've done and ask you to review that one. Then we'll be able to talk about fine points of those two approaches. I am more than happy to attribute both to you as you correctly noticed and corrected some issues with current code.

falkena:
@natdan If you mean me, than sure. Let's find the best solution for this.
If you not mean me, sorry about this comment.

natdan:
@falkena :( - that's what happens when you are trying to do 3 things at the same time. I am sorry for confusion. Yes, you :)

colinrgodsey:
Alright, I finally got this new dev board working yesterday. Had to kick the tires and fix a few userland drivers for some devices... but it does definitely work now. will verify tonight.

falkena:
@natdan , @colinrgodsey: Hi guys, any news for me? ;-)

natdan:
@falkena I am really sorry - I have many things put on hold until after 1st of April :(
(Taking kids from the game club to PiWars competition and failing to make our rover do simple follow the line. Eh... )
I'll pick it up immediately after. Apologies once more...

falkena:
@natdan: No problem. PiWars looks cool. Good luck to your kids!

colinrgodsey:
@falkena so sorry for the delay, i've also been having a hell of month too. Would you mind explaining the original issue to me? I actually have a branch (from master) ive been working on with this multiplexer and it seems to work fine. I haven't tested against vanilla master tho, but if you wouldn't mind, could you just go over the original issue for me again?

natdan:
@colinrgodsey I can pick it up now :) (we've ended up being 2nd overall!) I'll make another pull request so we can compare two approaches and find the better one... I am curious to hear what @falkena thinks of what I am going to do, too! :)

falkena:
@colinrgodsey It's really simple. Fix provided by @natdan for #303 creates a lot of /dev/i2c-??? entries,
independend, if PCA954 is connected or not. If PCA954 is not connected, all of them are dead, excepting /dev/i2c-1 on my Raspberry 2. The question i had was: "Can I avoid this creation?".  The answer was this pull-request :-)

natdan:
And I am happy with the idea, just tiny detail of implementation I would like discuss :) Good chance I have nothing tonight and I can go and do it already! :)

natdan:
@falkena Huh - I managed to do 'something'. Please have a look at: https://github.com/falkena/pi4j/pull/1
Check only changes in I2CProviderImpl - that's what I had in mind when I held up this pull request. Do let me know if you think that would work (or not and why!).

falkena:
Hi guys,
after a really crazy week, i reworked a pull-request a little bit :-)

pi4j-example/src/main/java/I2CExample.java
```diff
@@ -98,6 +99,26 @@ public static void main(String[] args) throws InterruptedException, PlatformAlre
...
+        
+        // find available busses
+        for (int number = I2CBus.BUS_0; number <= I2CBus.BUS_17; ++number) {
+            try {
+                I2CBus bus = I2CFactory.getInstance(number);
+                console.println("Supported I2C bus " + number + " found");
+            } catch (IOException exception) {
+                console.println("I/O error on I2C bus " + number + " occurred");
+            } catch (UnsupportedBusNumberException exception) {
+                console.println("Unsupported I2C bus " + number + " required");
```
colinrgodsey, line 118: ~~hmmm wont this print the error line for every unsupported bus? So if you just had the stock i2c-1 bus, you'd get 16 error messages for the others?~~

nvm, didnt notice what file it was

colinrgodsey:
is this updated with everything from your branch on the fork?

falkena:
@colinrgodsey I didn't merge from @natdan PR, since it was for discussion purpose only.
There are no new commits on pi4j:master, since my master is 10 commits ahead.

PR #335, opened 2017-04-23
IRQ Monitoring for PCF8574

gdepuille:
Hi,
I open this PR to adress an improvements to detect the modification on a GPIO Provider I2C (like PCF8574)

I have refactored and centralized the monitoring thread include in PCF8574GpioProvider and ADS1x15GpioProvider.
Now, like before, the thread check regularly if an input has changed. This will be detected when a native GPIO pin are in a specific state. 
In order to allow the possibility to use more than one device use the same IRQ input (PCF8574 has an open drain interrupt output), it's possible to disabled monitoring thread. 
I have implemented an exemple for the PCF8574 (I have it) to valid all the use case. But I haven't the ADS1x15, and I have not tested on this one.

When I realized some tests I have to meet a bug when the PCF8574 is used with inputs and outputs. The documentation says to define a line as an input a high level will be write to the single register. I have fix that with an output mask to store the input pins.

### DIFF
--- a/pi4j-core/pom.xml
+++ b/pi4j-core/pom.xml
@@ -15,6 +15,11 @@
 
 	<!-- PROJECT DEPENDENCIES -->
 	<dependencies>
+		<dependency>
+			<groupId>org.projectlombok</groupId>
+			<artifactId>lombok</artifactId>
+			<scope>provided</scope>
+		</dependency>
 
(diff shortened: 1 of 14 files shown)


FDelporte:
As this is an old PR, will close for now as the complete new V2 will allow extending in a different way.

PR #374, opened 2017-09-20
PiFace Control and Display LCD implementation

sbodmer:
The PiFace  Control and Display is a MCP23s17 and HD44780 device which is driven via SPI interface.
A new LCD implementation for the PiFace Control and Display (cad) board is available.
The LCD display and buttons can be driven with this class.


FDelporte:
Sorry, we will not be adding additional component support to pi4j v1 and this branch is in conflict state. We are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

PR #380, opened 2017-11-03
Newer and explicit licence-maven-plugin version

salk31:
Make build more stable by specifying licence-maven-plugin version and bumping current explicit version.

I was getting mangled source due to picking up random versions of the plugin. Making it match the other usage caused other problems so I've now pegged them both to the latest version.


salk31:
Sigh. I've muddled in another change. I'll try and tidy it up.

PR #390, opened 2017-11-22
Make build more stable by specifying licence-maven-plugin version and bumping current explicit version.

salk31:


salk31:
I was having trouble with 100s of source files being modified by the build making keeping track of what I'd actually changed a nightmare. I tracked it down to my build randomly picking up a version of the licence plugin. Specifying both of them and upgrading the plugin seemed to fix it.

PR #396, opened 2017-12-16
Avoid useless file changes after build

GeVanCo:
The template "template.ftl" causes useless file differences after a maven build, leading to all files having so-called changes.  This makes it difficult, if not impossible, to see the real and important file changes.
Template file is modified to avoid those useless changes in comment parts only.


GeVanCo:
Pls. change commit message from "stupid changes" to "Avoid useless file changes".  Was a bit pissed off when I saw that hunderds of files were changed after a fresh build and those changes were all nothing but comment changes due to a wrong template file...

Issue #323, opened 2017-03-02
SPI disabled on the RaspbPi


geohuk:
Hello,

I'm using pi4j-1.0.0 and I've got an issue with the PiFace initialization.

On the RaspberryPi if the SPI is disabled the com.pi4j.device.piface.impl.PiFaceDevice.PiFaceDevice() does not throw any exceptions but instead just prints "unable to open SPI device no such file or directory" and exits. This does not work for me, because the piface driver is just a module in my application and currently it crashes the whole application. 

Is there any better way to check if a PiFace is connected?

Cheers,
Georgi

PR #401, opened 2017-12-31
#323: Check if SPI device is available before try to open the device.

akuhtz:
This is a fix for the problem if the SPI is not enabled in the configuration. Found that #323 already exists and describes the symptoms I have.


FDelporte (approved):
Checked by Robert von Burg

PR #403, opened 2018-02-20
Support of Pimoroni Enviro pHat temperature sensor

aokunevich:


FDelporte:
Sorry, we will not be adding sensor support to pi4j v1, but are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

PR #405, opened 2018-03-12
Add GPIO PCF Analog Short (0-255) Event Dispatcher

zubial:
Add GPIO PCF Analog Short Input (0-255) Event Dispatcher

I tried it on https://github.com/zubial/homebrew-rasppi
It's a homebrewing system based on Rapsberry Pi 3

Best regards


FDelporte:
Sorry, we will not be adding additional features to pi4j v1, but are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

PR #444, opened 2019-03-03
Added support for MCP3202 ADC

bennymeg:
Added support for MCP3202 ADC including code examples.

Best,
Benny.


FDelporte:
Sorry, we will not be adding additional features pi4j v1, but are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

Issue #450, opened 2019-03-20
Add PIN list for different boards


FDelporte:
I have been experimenting with a JavaFX application to visualize the pin lay-out (and eventually later be able to show and change the GPIO remotely from this app to a board).

Based on the already existing enums I could quickly generate an image similar to the one on https://www.pi4j.com/1.2/pins/model-3b-plus-rev1.html, see Proof-Of-Concept code on https://github.com/FDelporte/pi4j-ui/blob/master/src/main/java/be/webtechie/pi4jui/board/impl/RaspberryPi_3B_Plus.java

Could this be something to add in pi4j itself?

(The dropdowns on the screenshot are just from experimenting, not related to the generated image yet)

![pi-pins](https://user-images.githubusercontent.com/1415873/54717455-7bef5b00-4b58-11e9-840c-5cc63aec2ce1.png)

savageautomate:
@FDelporte 

I do think something like this would be useful --- perhaps in an even greater capacity in version 2 where it could provide an interactive GUI to visualize the actual state of which pins are provisioned and what modes they are configured for as well as the active state.  Possibly even control of output states. 

The WiringPi project provides a GPIO utility to visualize the GPIO states via a command line interface.  

We might want to consider a runtime service in Pi4J v2 that exposes a remotely accessible interface for interactive with the I/O state.    This could coincide with the desire to also run a Pi4J enable project from your IDE/desktop and remote the I/O over the network to a physical Raspberry Pi.

PR #451, opened 2019-03-25
Develop/2.0.pinninginfo

FDelporte:
See https://github.com/Pi4J/pi4j/issues/450

My first pull request for this library, so guess it won't be first time right ;-)


FDelporte:
Will close this pull request for now as I want to extend it further. Will come back later ;-)

PR #456, opened 2019-04-07
Value 0 indicate success (return of ioctl)

williamchevremont:


FDelporte (approved):
Reviewed by Robert von Burg

PR #460, opened 2019-06-03
Develop/2.0.service

FDelporte:
Proposal for a Spring REST interface linked to the core library. Could be the starting point to also add websocket interface for more "realtime" remote interfacing.


FDelporte:
Do you want this REST service to be further developed in the 1.4-branch?

savageautomate:
Lets hold off.  1.4 may never be released -- its really a temporary placeholder for a JDK11 build.  However all efforts are really geared towards v2.0.   This type of service really belongs in 2.0 which will complement other v2.0 features.

Issue #453, opened 2019-04-02
PWM output check fails due to wrong operator


AbelPelser:
In [pi4j/pi4j-core/src/main/java/com/pi4j/io/gpio/GpioProviderBase.java:267,](https://github.com/Pi4J/pi4j/blob/03cacc62223cc59b3118bfcefadabab979fd84c7/pi4j-core/src/main/java/com/pi4j/io/gpio/GpioProviderBase.java#L267)
we believe that 
 `if (mode != PinMode.PWM_OUTPUT || mode != PinMode.SOFT_PWM_OUTPUT) {`
should be changed into 
 `if (mode != PinMode.PWM_OUTPUT && mode != PinMode.SOFT_PWM_OUTPUT) {`,

since mode can never be equal to both at the same time. This caused a test scenario to fail for us, halting our compilation process.

fclinton:
I added PR #465 I guess see if it gets merged or fork your own

savageautomate:
**Closed**.  Merged into v1.3 release.

PR #465, opened 2019-07-14
Incorrect logic checking for PWM Mode

fclinton:
Will always throw throw new InvalidPinModeException
Because to return false for mode != PinMode.PWM_OUTPUT || mode != PinMode.SOFT_PWM_OUTPUT
you would have to both be !=PinMode.PWM_OUTPUT and !=PinMode.SOFT_PWM_OUTPUT


fclinton:
related issue #453

PR #480, opened 2020-03-08
add synchronized block to pins iteration

ironman00023:
As java docs says on Collections.synchronizedList(List<T> list)
"It is imperative that the user manually synchronize on the returned list when iterating over it".


eitch:
Hi @ironman00023 nice catch! I've pushed this to my fork, where i have additional fixes for aarch64: https://github.com/eitch/pi4j/tree/develop/1.4

PR #487, opened 2020-04-21
Add support of Orange PI Zero platform

matyasb:


FDelporte:
Sorry, we will not be adding additional platform support to pi4j v1, but are working on a v2 which will be easier to extend with plugins. So this can be reviewed later.

PR #492, opened 2020-06-15
Add building instructions

eitch:


savageautomate:
I think this information is better suited for the existing BUILD INSTRUCTIONS file rather than the readme.  https://github.com/Pi4J/pi4j/blob/master/BUILD.INSTRUCTIONS

Thanks, Robert

PR #494, opened 2020-06-26
Added support for Docker Builders for Native Libraries (32-bit & 64-bit)

savageautomate:
**This change only applies to Pi4J v1.4.**

These changes add support for building the `pi4j-native` project (_native libraries_) using Docker images.   The `pi4j-native` project includes native (C) libraries that must be compiled for both CPU architectures `armhf` (32-bit) and `aarch64` (64-bit).  Building these native sources using the [Pi4J Docker Builder images](https://github.com/Pi4J/pi4j-docker) will simplify the build process by off-loading the build steps to containers that already have the proper ARM cross-compiler environments and toolchains established and configured.  

Additionally:
* Building the native libraries remotely via SSH/SCP has been removed.  
* Building the native libraries locally/directly on 32-bit Raspberry Pis [`armhf`] has been removed.
* Cross-compiling the native libraries locally/directly on 64-bit Raspberry Pis [`aarch64`] is supported. 
* Cross-compiling the native libraries locally/directly on 64-bit Linux [`amd64`] system is supported. 

**Note:**. It's important to note that the vast majority of users will never need to build the native libraries themselves. These libraries are pushed to Maven repositories as artifacts and can be downloaded just like other dependencies in a typical Maven build.  Only users who need to re-compile the native libraries will need to make use of either the Docker builders or cross-compiling.


savageautomate:
@eitch 

Are you interested in reviewing these changes before I merge into `develop/1.4`.

Thanks, Robert

eitch (approved):
I like what i see. I think this is a really good change.

PR #497, opened 2020-07-27
Removed exception handling (test smell)

eas5:
Problem:
The exception handling test smell occurs when a test method explicitly a passing or failing of a test method is dependent on the production method throwing an exception. 

Correction:
I used JUnit's exception handling to automatically pass/fail the test instead of writing custom exception handling code or throwing an exception. The original method generated two new ones: with and without the correct parameter, which must raise an exception.


FDelporte:
Thanks @eas5 for your first contribution! Having clear and consistent tests is indeed critical in a software project, so any support and improvement is very welcome!

PR #504, opened 2020-08-12
Develop/1.4

eitch:
This patch fixes the shutdown of the new single threaded GPIO event executor and also puts the GPIO incoming events as quickly as possible into their own thread, so that the native thread is blocked as short as possible.

Further the native thread only attaches once to the JVM, instead of once everytime an event arises. This should then create only one JVM thread per interrupt handling thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,12 @@ Copyright (C) 2012-2020 Pi4J
 
 ## Building
 
-To build locally without needing a Raspberry, simple cross-compile:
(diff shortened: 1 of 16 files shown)


eitch:
I have tested this on my production project and it works well.

eitch:
What is the status of this PR?

eitch:
@FDelporte @savageautomate any arguments against merging this PR?

